PR #25, opened 2021-08-30
Add minimal rustfmt.toml and .editorconfig

iljakuklic:
Here is a minimal `rustfmt.toml` and an `.editorconfig` to go with it.

Some options have been adjusted:
* `{chain,array}_width = 80` (up from 50, this is what Substrate does)
* `newline_style = "Unix"` (for consistency)
* `edition = "2018"`

Remaining options just state pre-existing defaults explicitly.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+[*]
+end_of_line = lf
+
+[*.rs]
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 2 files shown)


muursh (approved):

iljakuklic:
Removed a config option only available in nightly. Also checked that the stable and nightly rustfmt give the same result on the main repo and on chainscript repo.

altonen (approved):
LGTM

PR #28, opened 2021-09-01
Update build.yml

muursh:
Run cargo check on macos, windows and linux for both staging and master PRs.
Only run full build and tests on master PR


altonen (approved):
Looks good to me!

PR #31, opened 2021-09-06
Apply rustfmt for all project code.

sinitcin:
As we discussed before that we shall use the rustfmt. In this commit  just results of executed:

`cargo fmt --all `


muursh (approved):
I'm okay with this. We said we'd do it so a separate pr makes sense. CI passes, builds fine locally and there's nothing weird so I'm all good

sinitcin:
@muursh Thank you!

PR #33, opened 2021-09-08
UTXO: Adapt input and output formats for more flexibility

iljakuklic:
This changes the format of transaction inputs and outputs to enable more
future extensibility. For outputs, the `Destination` type is introduced
to differentiate between various spending types such as P2SH, P2PK,
P2PKH and so on. It is an enum so it is easily extended by adding extra
variants. In inputs, there are two new fields: `lock` and `witness`.
These are binary data that are interpreted according to the output
destination.

TODO: Update `utxo-tester` to reflect the changes introduced here.


altonen (approved):
LGTM. My only concern is how these new enum values are used with python so I'll be postponing merging this to my local node until I'm done with PP tests. Previously the values were used as`u8`. But looks good and I think you should add yourself to the authors section of `lib.rs`.

muursh:
Agreed that you should add yourself to the authors at this point

sinitcin:
Perfect Lukas!

muursh (approved):

sinitcin (approved):

iljakuklic:
> LGTM. My only concern is how these new enum values are used with python so I'll be postponing merging this to my local node until I'm done with PP tests. Previously the values were used as`u8`. But looks good and I think you should add yourself to the authors section of `lib.rs`.

I am in the process of updating the `utxo-tester` to reflect the changes introduced here. I don't intend to merge this until it's ready and then merge both at the same time.

pallets/utxo/src/lib.rs
```diff
@@ -320,25 +394,24 @@ pub mod pallet {
...
+                Destination::CreatePP(_, _) => {
                     log::info!("TODO validate OP_CREATE");
                 }
-                ScriptType::OpCall => {
+                Destination::CallPP(_, _) => {
                     log::info!("TODO validate OP_CALL");
                 }
             }
+            // checked add bug in example cod where it uses checked_sub
+            total_output =
```
altonen, line 413: Extremely minor thing but running cargo-fmt told that this line is not formatted correctly. If you're doing any other changes to this commit anymore, the code could be formatted.

iljakuklic: Cool, will change that, thanks for pointing this out

iljakuklic:
PR for `utxo-tester` is up https://github.com/mintlayer/utxo-tester/pull/2

PR #35, opened 2021-09-10
Fix transaction output indexing

iljakuklic:
The transactions have been indexed in an inconsistent way signalling
potential problems with pending transactions and with getting
appropriate UTXO hashes.


muursh (approved):
Looks good to me.

PR #40, opened 2021-09-14
Update build.yml

muursh:


muursh:
Windows build fails but that seems unavoidable right now. 
I've created a branch to handle a windows build with the fixes necessary but haven't had time to get it running yet

altonen (approved):
Why do you want to merge this to staging even if it doesn't work? I don't see the point of it, the update itself looks fine

PR #42, opened 2021-09-22
Basic scripting integration

iljakuklic:
Not particuralry thoroughly tested for now. The test present is more of an usage example. The way transactions are signed is about to be changed. For now, just sign the transaction with all witness fields set to empty (`get_simple_transaction` does this). Not everything is implemented but most common features are working. Some resource limits are not enforced yet, time locks are not implemented, Mintlayer-specific extensions and some other minor things missing as well.

### DIFF
--- a/pallets/utxo/src/script.rs
+++ b/pallets/utxo/src/script.rs
@@ -1,0 +1,150 @@
+// Copyright (c) 2021 RBB S.r.l
+// opensource@mintlayer.org
+// SPDX-License-Identifier: MIT
+// Licensed under the MIT License;
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://spdx.org/licenses/MIT
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Author(s): L. Kuklinek
+
+use codec::DecodeAll;
+use codec::{Decode, Encode};
+use core::convert::TryInto;
+use frame_support::sp_io::crypto;
+use sp_core::sr25519;
+use sp_std::prelude::*;
+use variant_count::VariantCount;
+
+/// An unvalidated signature type
+type RawSignatureType = u8;
+
+/// A signature together with its usage information
+struct Signature {
+    /// The raw signature data.
+    signature: Vec<u8>,
+    /// Sighash specifies which parts of the transaction are included in the hash that is verified
+    /// by the signature.
+    /// TODO: Sighash is not implemented at the moment, `SIGHASH_ALL` mode is used for everything.
+    sighash: (),
+}
+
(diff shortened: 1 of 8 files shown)


pallets/utxo/src/script.rs
```diff
@@ -0,0 +1,150 @@
...
+        let Signature {
+            signature: sig,
+            sighash: _sighash,
+        } = sig;
+        match pk {
+            Ok(Public::Schnorr(pk)) => (&sig[..]).try_into().map_or(false, |sig| {
+                crypto::sr25519_verify(&sr25519::Signature::from_raw(sig), self.tx_data, pk)
+            }),
+            // Unrecognized signature type => accept the signature.
+            Err(_pk) => true,
```
altonen, line 123: I was just wondering, would it pose any problems if I would submit TXs with unrecognized but also invalid signatures and they were accepted and included in the blockchain (would they be included?) and it was only later found out when Mintlayer would adopt this new signature scheme that the blockchain contains TXs with invalid signatures. Can this "attack" actually happen solely because we're returning `true` here and if so, would it be an actual problem?

altonen (approved):
LGTM

pallets/utxo/src/script.rs
iljakuklic: When new signature types are introduced while blockchain is live, we can only start enforcing the new rules starting at some block height in the future in order to not compromise the existing chain. Bitcoin has another layer of protection against the sort of thing you describe. If a transaction looks like using something reserved for future extensions, the node does not forward it. These are so-called "non-standard" transactions. However, if such a transaction ends up in a block, it is deemed valid according to the consensus rules. So I think we should at some point have a way of differentiating between valid, non-standard and invalid transactions/blocks.

iljakuklic: I was thinking a bit about how to do the softforing updates without reimplementing Substrate. One fairly generic idea could be to use `enum`s where extra variants could be added as the system is developed. For example, in case of Singatures, it could look something like this:

```rust
#[derive(Encode, Decode, Softfork, ...)]
enum Pubkey {
    // Schnorr has been here since the beginning
    Schnorr([u8; 32]),

    // BLS was introduced starting block 1234
    #[softfork(since = 1234)]
    BLS([u8; 32]),
}
```

Then there would be an `Extensible<T>` type that wraps enums like this so that Substrate does not kick out things it is not (yet) able to decode. The `Encode` and `Decode` traits would have to be implemened manually for the `Extensible` type to get desired behaviour. Usage example:
```rust
    let pk: Extensible<Pubkey> = todo!("something automatically decoded by substrate");
    match pk {
        Extensible::Ok(pubkey) => todo!("do something with pubkey of type Pubkey"),
        Extensible::Unrecognized(_) => todo!("handle possible future extension"),
    }
```

Thoughts and/or comments?

muursh: This might need some thinking about when it comes to the CTs if we go with some ZK stuff as they won't slot nicely in to this architecture but 1) that's years away and 2) that's a me problem

muursh (approved):

pallets/utxo/src/script.rs
muursh: I like that idea a lot

PR #43, opened 2021-09-24
Bootnodes

altonen:
If `--dev` is used when the node is started, the bootnode list is not fetched to prevent unnecessary interference with testing. Node also doesn't establish a permanent connection with the hard-coded nodes listed in the chainspec because they're running a different chainspec than dev. Thus, this change should not interfere with anybody's development work.

If you want to test this, you can find a list of bootnodes from `https://raw.githubusercontent.com/altonen/core/master/assets/bootnodes.json`

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -2,13 +2,13 @@ use node_template_runtime::{
     pallet_utxo, AccountId, AuraConfig, BalancesConfig, GenesisConfig, GrandpaConfig, PpConfig,
     Signature, SudoConfig, SystemConfig, UtxoConfig, WASM_BINARY,
 };
+use sc_network::config::MultiaddrWithPeerId;
 use sc_service::ChainType;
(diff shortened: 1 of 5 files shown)


node/src/chain_spec.rs
```diff
@@ -2,14 +2,14 @@ use node_template_runtime::{
     pallet_utxo, AccountId, AuraConfig, BalancesConfig, GenesisConfig, GrandpaConfig, PpConfig,
     Signature, SudoConfig, SystemConfig, UtxoConfig, WASM_BINARY,
 };
+use sc_network::config::MultiaddrWithPeerId;
 use sc_service::ChainType;
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
+use sp_core::H256;
```
muursh, line 8: I got really confused as to why you were adding this for a second until I double checked and saw you removing it just below

muursh (approved):

PR #44, opened 2021-09-27
The basic implementation of tokens on utxo 

sinitcin:
This basic token system allows to create, send and view a list of tokens.

Execute the extrinsic `tokensCreate` for creating new tokens. It allows making a new transaction with an output that includes a new tokens amount:
```
utxo -> tokensCreate

- pubkey of the creator
- An input for taking a creation fee. 
- tokenName
- tokenTicker or short name
- supply amount
```

For sending tokens it's able to use `spend` just check TokenId in TxOutputHeaders.

For request a list of tokens: 
```
curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{
    "jsonrpc":"2.0",
    "id":1,
    "method":"tokens_list",
    "params": []
}'
```
Returns the array with TokenIds and their names:
```js
[
    {1, "My test token"},
    {2, "Global universe super token"},
]
```


types.js
```diff
@@ -0,0 +1,52 @@
...
+  "String": "Vec<u8>",
+  "TokenID": "u64",
+  "TokenInstance": {
+    "id": "u64",
+    "name": "String",
+    "ticker": "String",
+    "supply": "u128"
+  },
+  "TokenListData": "Vec<TokenInstance>"
+}
```
iljakuklic, line 52: Good idea to put this into a file

pallets/utxo/src/lib.rs
```diff
@@ -286,6 +322,14 @@ pub mod pallet {
     #[allow(type_alias_bounds)]
     pub type TransactionFor<T: Config> = Transaction<T::AccountId>;
 
+    #[pallet::storage]
+    #[pallet::getter(fn token_list)]
+    pub(super) type TokenList<T> = StorageValue<_, TokenListData, ValueQuery>;
```
iljakuklic, line 327: Two general comments/question about `TokenList`:

1. Could this be a `StaorageMap` taking token ID to data. Should be faster when looking up token based on ID
2. Do we limit token name somehow? It is stored by all nodes and has to be in some sort of hot storage so an adversary may try to put some very big thing there.

iljakuklic (commented):
Looks good except the concerns about `.unwrap()` I commented on which is the only thing I would consider a blocker. If that is resolved or explained, this is fine.

pallets/utxo/src/header.rs
```diff
@@ -59,142 +213,120 @@ u16_to_enum! {
...
-
-    pub(crate) fn insert(header: &mut TXOutputHeader, signature_method: SignatureMethod) {
-        *header = header.clone() & 0b1_111111_111111_000; // remove the original signature, if any.
-        let signature_method = signature_method as u16;
-        *header = header.clone() | signature_method;
+u128_to_enum! {
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Clone, Encode, Decode, Eq, PartialEq, PartialOrd, Ord, Hash, Debug)]
+pub enum TokenType {
+    MLT = 0,
```
muursh, line 220: MLT is technically not a MLS-01 token so does it need a token ID? If we remove it how can we distinguish MLS-01 and MLT?

muursh (commented):
I have no real issues with this bar the whole MLT/MLS-01 token ID thing which I'm not sure is a real issue.

Token IDs need to be random and we need to ensure a token doesn't exist with the same ID I guess? Or do we chance it?

pallets/utxo/src/lib.rs
muursh: Agreed we should have a size limit here.

pallets/utxo/Cargo.toml
```diff
@@ -21,6 +21,8 @@ std = [
 hex-literal = "0.2.1"
 log = "0.4.8"
 serde = '1.0.119'
+getrandom = { version = "0.2", features = ["js"] }
```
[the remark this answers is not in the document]
sinitcin: Yeah, good catch! I wished to use Strings in UTXO for tokens names and tickers. In theory, I shall add the `std` feature for all dependencies and `getrandom` in the UTXO manifest but I had problems with them. Some of these in std had different dependencies for the same crates in Substrate code and I couldn't resolve it.

pallets/utxo/rpc/src/lib.rs
```diff
@@ -26,6 +27,13 @@ use std::sync::Arc;
 pub trait UtxoApi<BlockHash> {
     #[rpc(name = "utxo_send")]
     fn send(&self, at: Option<BlockHash>) -> Result<u32>;
+
+    // What means Vec<(u64, Vec<u8>)> ?
+    // At the moment we have some problems with use serde in RPC, we can serialize and deserialize
+    // only simple types. This approach allow us to return Vec<(TokenId, TokenName)> instead of
+    // pallet_utxo_tokens::TokenListData
+    #[rpc(name = "tokens_list")]
+    fn tokens_list(&self, at: Option<BlockHash>) -> Result<Vec<(u64, Vec<u8>)>>;
```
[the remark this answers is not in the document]
sinitcin: Sure, I'll do

pallets/utxo/src/lib.rs
```diff
@@ -82,6 +117,7 @@ pub mod pallet {
 
     pub trait WeightInfo {
         fn spend(u: u32) -> Weight;
+        fn tokens_create(u: u32) -> Weight;
```
altonen, line 120: This is minor thing but shouldn't this be `token_create`?

pallets/utxo/src/lib.rs
```diff
@@ -297,7 +341,10 @@ pub mod pallet {
 
     #[pallet::event]
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
+    #[pallet::metadata(T::AccountId = "AccountId", T::AssetId = "AssetId")]
```
altonen, line 344: Is `AssetId` needed here? At least `TokenCreated` below doesn't use it.

pallets/utxo/src/lib.rs
```diff
@@ -562,6 +677,58 @@ pub mod pallet {
...
+        token_ticker: String,
+        supply: Value,
+    ) -> DispatchResultWithPostInfo {
+        ensure!(!supply.is_zero(), Error::<T>::MinBalanceZero);
+
+        // Take a free TokenID
+        let token_id =
+            <TokensHigherID<T>>::get().checked_add(1).ok_or("All tokens IDs has taken")?;
+        sp_runtime::print("TOKEN ID IS");
+        sp_runtime::print(&token_id);
```
altonen, line 694: It may be easier to use `log::info!` here. I don't know how `sp_runtime::print` works but it could be written like this instead: `log::info!("TOKEN ID IS {:?}", token_id);`

pallets/utxo/src/lib.rs
```diff
@@ -573,6 +740,28 @@ pub mod pallet {
...
+            let caller = &ensure_signed(origin)?;
+            tokens_create::<T>(
+                caller,
+                public,
+                input_for_fee,
+                token_name,
+                token_ticker,
+                supply,
+            )?;
+            Self::deposit_event(Event::<T>::TokenCreated(1u32, caller.clone()));
```
altonen, line 762: what does the `1u32` denote?

pallets/utxo/src/lib.rs
```diff
@@ -408,8 +454,42 @@ pub mod pallet {
...
+        let tokens_list = <TokenList<T>>::get();
+        for output in tx.outputs.iter() {
+            let tid = OutputHeaderData::new(output.header).token_id();
+            // If we have input and output for the same token it's not a problem
+            if full_inputs.iter().find(|&x| (x.0 == tid) && (x.1 != *output)).is_some() {
+                continue;
+            } else {
+                // But when we don't have an input for token but token id exist in TokenList
+                ensure!(
+                    tokens_list.iter().find(|&x| x.id == tid).is_none(),
```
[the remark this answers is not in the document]
altonen: nvm this was pointed out by @iljakuklic already

pallets/utxo/src/header.rs
```diff
@@ -13,19 +13,173 @@
...
+}
+
+// Size of bit fields, total 72 bits
+const SIGNATURE_METHOD_SIZE: u32 = 3;
+const TOKEN_ID_SIZE: u32 = 64;
+const VERSION_SIZE: u32 = 5;
+
+#[derive(Debug)]
+pub struct OutputHeaderData {
+    sign_method: BitsField,
```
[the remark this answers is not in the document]
sinitcin: Agree with you, I'll remove that from `OutputHeaderData`. I didn't touch this place before because thought it needs for some reason

pallets/utxo/src/lib.rs
```diff
@@ -562,6 +677,58 @@ pub mod pallet {
...
+
+        // Take a free TokenID
+        let token_id =
+            <TokensHigherID<T>>::get().checked_add(1).ok_or("All tokens IDs has taken")?;
+        sp_runtime::print("TOKEN ID IS");
+        sp_runtime::print(&token_id);
+
+        // Input with MLT FEE
+        let fee = UtxoStore::<T>::get(input_for_fee.outpoint).unwrap().value;
+        ensure!(fee < 99, Error::<T>::Unapproved);
```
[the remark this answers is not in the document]
sinitcin: So, about the first point, I read somewhere that `unwrap` in the Substrate does not make panics, correct me if I'm wrong. Nevertheless, we can change this place for a more comfortable view =)
```rust
let fee = UtxoStore::<T>::get(input_for_fee.outpoint).ok_or(Error::<T>::Unapproved)?.value;
```
The second point, I've double-checked it and you are right!

types.js
sinitcin: Yep, I think we need one common place for js

pallets/utxo/src/lib.rs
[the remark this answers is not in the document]
iljakuklic: Yes I think `.ok_or()` is preferable. Regarding `unwrap` not panicking, I did not find `Option` docs for substrate, they probably just use `Option` from Rust core. The substrate doc for `Result` suggest otherwise: https://substrate.dev/rustdocs/latest/sp_std/result/enum.Result.html#method.unwrap

pallets/utxo/rpc/runtime-api/src/lib.rs
```diff
@@ -13,11 +13,18 @@
...
+use frame_support::inherent::Vec;
+
 sp_api::decl_runtime_apis! {
     pub trait UtxoApi {
         fn send() -> u32;
+        // What means Vec<(u64, Vec<u8>)> ?
+        // At the moment we have some problems with use serde in RPC, we can serialize and deserialize
+        // only simple types. This approach allow us to return Vec<(TokenId, TokenName)> instead of
+        // pallet_utxo_tokens::TokenListData
+        fn tokens_list() -> Vec<(u64, Vec<u8>)>;
```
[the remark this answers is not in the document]
sinitcin: > Do we have numbers for how slow this gets with a massive token list?

So, I'll prepare it 

> Token IDs need to be random
Did you mean that incrementing the `TokensHigherID` isn't suited for us?

pallets/utxo/src/lib.rs
muursh: I'm not sure we even need it that big. 5 for the ticker and 25 for the token. Going through the 100 largest coins by market cap every single one will fit in 5/25

pallets/utxo/src/header.rs
```diff
@@ -13,19 +13,173 @@
...
+struct BitsField {
+    length: u32,
+    offset: u32,
+    pub data: u128,
+}
+
+// Size of bit fields, total 72 bits
+const SIGNATURE_METHOD_SIZE: u32 = 3;
+const TOKEN_ID_SIZE: u32 = 64;
+const VERSION_SIZE: u32 = 5;
```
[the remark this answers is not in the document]
sinitcin: > So I guess we could do with a token type here 

Absolutely agree with you!

pallets/utxo/src/lib.rs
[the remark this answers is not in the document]
muursh: Okay I've just noticed something else. The 100 tokens here is in MLT which means this is the base unit (the sat of the MLT world), we really need a name for this base unit.

https://github.com/mintlayer/core/blob/36599cea0eafbac785a89ced09f0fdb3233b1042/node/src/chain_spec.rs#L150
You can see here in the genesis block that an MLT is defined as 1_000 * 100_000_000 base units so we don't have to worry about floating point stuff because that would give me nightmares.

This 100 should be 100 * 1_000 * 100_000_000

@anyxem we need to sort the UI so it understands this too. 1 unit is not 1 MLT just like 1 sat is not 1 btc.

pallets/utxo/src/header.rs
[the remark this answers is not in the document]
sinitcin: Done

pallets/utxo/src/header.rs
sinitcin: In your case, you can check `TokenType`. Now `TokenID` and `TokenType` are different things.

pallets/utxo/src/lib.rs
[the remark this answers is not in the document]
sinitcin: > we really need a name for this base unit.

Proposal - 100 * 1_000 * 100_000_000 Mint seeds == 100 MLT?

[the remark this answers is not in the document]
sinitcin: The default weight for a token tx is fixed

pallets/utxo/src/lib.rs
sinitcin: Good question. @muursh is there we create token or tokenS? For example, we create 1000 tokens with a new token ID. Help us with proper English please =)

muursh: Okay I say the singular token makes more sense here.

I think token_create is the better wording

sinitcin: Done

pallets/utxo/src/lib.rs
```diff
@@ -573,6 +740,28 @@ pub mod pallet {
             Self::deposit_event(Event::<T>::TransactionSuccess(tx));
             Ok(().into())
         }
+
+        #[pallet::weight(T::WeightInfo::tokens_create(768_usize.saturating_add(token_name.len()) as u32))]
+        pub fn tokens_create(
```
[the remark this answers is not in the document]
sinitcin: Done

pallets/utxo/src/lib.rs
```diff
@@ -219,6 +255,17 @@ pub mod pallet {
             }
         }
 
+        pub fn new_tokens(token_id: TokenID, value: Value, pub_key: H256) -> Self {
```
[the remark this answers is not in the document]
sinitcin: Fixed

pallets/utxo/src/lib.rs
sinitcin: Yep, the good catch. At the moment it's no need for us.

sinitcin: But for the further tokens, it will.

sinitcin: Removed

pallets/utxo/src/lib.rs
sinitcin: Thank you for pointing, this print I used to debug, and I thought that I've already removed it. This is an unnecessary print here.

sinitcin: Removed now

pallets/utxo/src/lib.rs
sinitcin: > I'm not sure we even need it that big. 5 for the ticker and 25 for the token. Going through the 100 largest coins by market cap every single one will fit in 5/25

Good!

sinitcin: Done!

pallets/utxo/src/lib.rs
sinitcin: Another good catch, it should be TokenID

pallets/utxo/rpc/runtime-api/src/lib.rs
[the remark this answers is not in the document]
sinitcin: So, for 1000 insertions: 
```
test result: ok. 1 passed; finished in 1.21s
```

Unfortunately, for 10_000 insertions:
```
test result: ok. 1 passed; finished in 102.43s
```

And reading all token_list for 1000 times:
```
test result: ok. 1 passed; finished in 167.19s
```
Unfortunately, for 10_000 insertions was too long =(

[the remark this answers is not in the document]
sinitcin: @muursh

pallets/utxo/src/lib.rs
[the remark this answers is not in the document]
sinitcin: As we mention today at the meeting for the test net it's enough.

pallets/utxo/src/lib.rs
sinitcin: Fixed

pallets/utxo/src/lib.rs
```diff
@@ -678,21 +679,21 @@ pub mod pallet {
...
-    pub fn tokens_create<T: Config>(
+    pub fn token_create<T: Config>(
         caller: &T::AccountId,
         public: H256,
         input_for_fee: TransactionInput,
         token_name: String,
         token_ticker: String,
         supply: Value,
-    ) -> DispatchResultWithPostInfo {
+    ) -> Result<u64, DispatchErrorWithPostInfo<PostDispatchInfo>> {
```
muursh, line 689: :)

pallets/utxo/src/tests.rs
```diff
@@ -400,3 +400,36 @@ fn test_tokens() {
         assert_ok!(Utxo::spend(Origin::signed(0), tx.clone()));
     });
 }
+    
+#[test]
+fn attack_double_spend_by_tweaking_input() {
```
muursh, line 405: This is @iljakuklic 's code right? The stuff that's just been merged? It looks identical but there might be a slight difference I'm missing

iljakuklic: There is a commit that merges in `staging` but I am not sure why this shows up in the diff

pallets/utxo/src/lib.rs
```diff
@@ -563,6 +679,58 @@ pub mod pallet {
...
+        ensure!(token_ticker.len() <= 5, Error::<T>::Unapproved);
+        ensure!(!supply.is_zero(), Error::<T>::MinBalanceZero);
+
+        // Take a free TokenID
+        let token_id =
+            <TokensHigherID<T>>::get().checked_add(1).ok_or("All tokens IDs has taken")?;
+
+        // Input with MLT FEE
+        let fee = UtxoStore::<T>::get(input_for_fee.outpoint).ok_or(Error::<T>::Unapproved)?.value;
+        ensure!(fee >= 100, Error::<T>::Unapproved);
```
muursh, line 700: This still needs changing. The fee should be >= 100 * 1_000 * 100_000_000

pallets/utxo/src/lib.rs
```diff
@@ -563,6 +679,58 @@ pub mod pallet {
...
+                input_for_fee,
+            ],
+            outputs: crate::vec![
+                // Output a new tokens
+                TransactionOutput::new_token(token_id, supply, public),
+            ],
+        };
+
+        // We shall make an output to return odd funds
+        if fee > 100 {
```
muursh, line 716: Again this should be 100 * 1_000 * 100_000_000

pallets/utxo/src/lib.rs
```diff
@@ -563,6 +679,58 @@ pub mod pallet {
...
+            tx.outputs.push(TransactionOutput::new_pubkey(fee - 100, public));
```
muursh, line 717: and him

pallets/utxo/src/weights.rs
```diff
@@ -40,4 +40,9 @@ impl<T: frame_system::Config> crate::WeightInfo for WeightInfo<T> {
             .saturating_add(T::DbWeight::get().reads(3 as Weight))
             .saturating_add(T::DbWeight::get().writes(3 as Weight))
     }
+
+    fn token_create(u: u32) -> Weight {
+        // Under construction
+        (u as Weight).saturating_add((100 * 1_000 * 100_000_000 as Weight))
```
muursh, line 46: Ah this is where the confusion has come from with the 100 vs 100 * 1_000 * 100_000_000 the fee should be the big one. I think we can leave the weight as it was before

pallets/utxo/src/lib.rs
muursh: Okay let me explain

iljakuklic: This deserves a constant. Or, to be even more defensive, a special type (idk maybe `MTLAmount`) with two constructors like `from_mlt` and `from_mintseed` so we are explicit about the unit used everywhere.

muursh: 1 MLT = 1 * 1_000 * 100_000_000 base units, let's call them pennies.
Because we need MLT to be divisible and we don't handle floating point numbers the base unit, the penny, is the smallest division of the MLT. 

If a user sends 1 MLT they actually send 1 * 1_000 * 100_000_000 base units/pennies. To send 500 MLT they'd send 500 * 1_000 * 100_000_000 base units/pennies

The 100 unit fee here is actually 100/(1_000 * 100_000_000) which is what a nano-MLT...

Think of bitcoin, there's no such thing as a bitcoin in the code base, everything is in sats and the idea of a bitcoin is essentially a UI idea. https://www.investopedia.com/terms/s/satoshi.asp

muursh: https://github.com/mintlayer/core/blob/36599cea0eafbac785a89ced09f0fdb3233b1042/node/src/chain_spec.rs#L150

Check here where we create 400_000_000 coins in the genesis block

sinitcin: Fixed

pallets/utxo/src/lib.rs
sinitcin: Fixed

pallets/utxo/rpc/runtime-api/src/lib.rs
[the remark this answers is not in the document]
muursh: I mean 10k insertions is slow. But also never going to happen in one go I guess. I think we're good for the testnet but let's rethink this after

pallets/utxo/src/weights.rs
sinitcin: Would it be correct? 

```rust
(u as Weight).saturating_add((100 as Weight))
```

pallets/utxo/src/tests.rs
sinitcin: @muursh I've just merged it from staging. Should be correct but let @iljakuklic recheck that.

pallets/utxo/src/weights.rs
muursh: Yep I think so, the weight is the tx weight rather than the fee so there's no need for it to be massive

pallets/utxo/src/tests.rs
sinitcin: > There is a commit that merges in staging but I am not sure why this shows up in the diff

Okay, if it is correct, let's resolve this

pallets/utxo/src/weights.rs
sinitcin: Done

pallets/utxo/rpc/runtime-api/src/lib.rs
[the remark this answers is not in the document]
sinitcin: @muursh sure

muursh (approved):
Sorry about the confusion here on the weight.

LGTM. If nobody else has concerns then we're good to go

iljakuklic (approved):

sinitcin:
Closed due to unverified commit

PR #47, opened 2021-09-30
Bech32 support

altonen:
This PR implements Bech32 support to the UTXO system. It firstly introduces a layer of abstraction by providing a function `send_to_address()` which can also be found from Bitcoin's RCP API. This function takes the origin, value and destination address as input and converts that to a UTXO transaction. As Mintlayer doesn't support SegWit, normal P2PKH is used instead.

This implementation has two major limitations I'd like your comments on:

- In its current form, `send_to_address` is only able to use `Destination::Pubkey` as input since the owner of those inputs can be trivially found out. Optimally our wallet would contain a map of "our UTXO" where we could easily select UTXO for spending but in current development environment I'm not sure that's possible because there is no single user or wallet AFAICT.
- Because I believe the key management is still an open issue and there seems to be no active key context when the node is run, `send_to_address()` currently only works in unit tests. I know this is a major limitation but changing the way it works either requires changing the design which I'm open to or waiting for (I believe Ben's?) key management updates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -409,6 +409,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
+name = "bech32"
+version = "0.1.0"
+
+[[package]]
 name = "bincode"
 version = "1.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


muursh:
Yep I'm doing 2 things that will have some impact on this, 1 the sig stuff which will be done soon and 2 tearing apart the km stuff which won't be done quite so soon because whenever I try and do anything with it something else pops up :)

sinitcin (approved):
Good!

muursh (approved):
LGTM.

PR #49, opened 2021-09-30
Transaction signatures

iljakuklic:
Update the transaction signature mechanism. It is a kind of hybrid between Segwit and Taproot way of doing signatures. Since generating signatures is no longer so straightforward, there will be matching utxo-tester PR soon which should help with experimenting and as a second reference implementation.

### DIFF
--- a/pallets/utxo/src/script.rs
+++ b/pallets/utxo/src/script.rs
@@ -15,75 +15,38 @@
 //
 // Author(s): L. Kuklinek
 
-use codec::DecodeAll;
-use codec::{Decode, Encode};
-use core::convert::TryInto;
-use frame_support::sp_io::crypto;
-use sp_core::sr25519;
+use crate::{sign, Transaction, TransactionOutput};
+use chainscript::context::ParseResult;
+use codec::Encode;
 use sp_std::prelude::*;
-use variant_count::VariantCount;
-
-/// An unvalidated signature type
-type RawSignatureType = u8;
-
-/// A signature together with its usage information
-struct Signature {
-    /// The raw signature data.
-    signature: Vec<u8>,
-    /// Sighash specifies which parts of the transaction are included in the hash that is verified
-    /// by the signature.
-    /// TODO: Sighash is not implemented at the moment, `SIGHASH_ALL` mode is used for everything.
-    sighash: (),
-}
-
-/// A public key. An enum to accommodate for multiple signature schemes.
-#[repr(u8)]
-#[derive(Eq, PartialEq, Clone, Copy, Encode, Decode, Debug, VariantCount)]
-enum Public {
-    /// Schnorr public key
-    Schnorr(sr25519::Public),
-}
 
 /// Mintlayer script context.
-struct MLContext<'a> {
-    tx_data: &'a [u8],
+struct MLContext<'a, AccountId> {
+    tx: &'a Transaction<AccountId>,
+    utxos: &'a [TransactionOutput<AccountId>],
+    index: u64,
 }
 
-impl<'a> MLContext<'a> {
-    fn new(tx_data: &'a [u8]) -> Self {
-        MLContext { tx_data }
-    }
-}
-
-impl chainscript::Context for MLContext<'_> {
-    /// Maximum number of bytes pushable to the stack
-    ///
-    /// We do not limit the size of data pushed into the stack in Mintlayer since programmable pool
-    /// binaries may be fairly large. Maximum size is still subject to `MAX_SCRIPT_SIZE`.
-    const MAX_SCRIPT_ELEMENT_SIZE: usize = usize::MAX;
-
-    /// Maximum number of public keys per multisig
-    const MAX_PUBKEYS_PER_MULTISIG: usize = 20;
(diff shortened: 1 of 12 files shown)


muursh:
Looks like @altonen got in before the conflicts :smile:

iljakuklic:
> Looks like @altonen got in before the conflicts smile

Yea, I kind of expected there to be conflicts, will resolve them soon.

muursh:
They're pretty minor

iljakuklic:
Rebased and rolling again :]

muursh:
Do you think it's worth generalising these H256 and H512 datatypes littered around the place to Vec<u8>? That way we never need to worry about pub key/sig/address/anything size it will always be supported no matter what

iljakuklic:
> So the reason I'm thinking about this is BLS will need new data types as it won't fit in 256/512 bit datatypes, the CT stuff will certainly need different datatypes. Now we can define H384 in substrate itself and handle it that way or we can stick the Vec. I do get where you're coming from with regards to the compiler though.

My preference would be `H384` or even better something like `bls::Public` and `bls::Signature`. If you take a look into `sign.rs` in this PR at the `Signature` type there, it is an enum exactly for the reason that we will want to add more variants in there. All of this will probably need some amount of redesign for signature batching but I did not want to engage in premature design for that. Does that sound sensible?

muursh:
Alrighty... So i have to admit I hadn't seen the sign.rs file yet. You can ignore my original comments having had a quick look.
I have already defined bls::Public and bls::Signature in an identical way to sr25519 so this should drop in nicely. I'll take a proper look through the PR in the morning

Yeah we've got a few things that ultimately we'll need to support, we have the code for multisig, threshold sigs and aggregated sigs already but we won't implement it until after the testnet.

muursh:
Alrighty... So i have to admit I hadn't seen the sign.rs file yet. You can ignore my original comments having had a quick look.
I have already defined bls::Public and bls::Signature in an identical way to sr25519 so this should drop in nicely. I'll take a proper look through the PR in the morning

Yeah we've got a few things that ultimately we'll need to support, we have the code for multisig, threshold sigs and aggregated sigs already but we won't implement it until after the testnet.

iljakuklic:
More conflicts, yay! Will resolve soon but after that `pallets/utxo/src/lib.rs` will have chenged.

pallets/utxo/src/script.rs
```diff
@@ -95,56 +58,57 @@ impl chainscript::Context for MLContext<'_> {
...
-            sighash: _sighash,
-        } = sig;
-        match pk {
-            Ok(Public::Schnorr(pk)) => (&sig[..]).try_into().map_or(false, |sig| {
-                crypto::sr25519_verify(&sr25519::Signature::from_raw(sig), self.tx_data, pk)
-            }),
-            // Unrecognized signature type => accept the signature.
-            Err(_pk) => true,
-        }
+    fn verify_signature(&self, sd: &Self::SignatureData, _: &[u8], sep_idx: u32) -> bool {
```
altonen, line 71: What is the third parameter and why is it not used?

pallets/utxo/src/sign.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub fn construct<AcctId: Encode>(
+        sighash: SigHash,
+        tx: &Transaction<AcctId>,
+        spending: &[TransactionOutput<AcctId>],
+        index: u64,
+        codesep_idx: u32,
+    ) -> Self {
+        let idx = index as usize;
+        assert!(spending.len() == tx.inputs.len());
+        assert!(idx < tx.inputs.len());
```
altonen, line 103: Are these intentionally `assert` instead of `ensure`?

altonen (commented):
LGTM but the conflicts need to be resolved and cargo said there was some unformatted code, can you check it on your computer?

pallets/utxo/src/sign.rs
iljakuklic: Yes, it is the responsibility of the caller to ensure the index refers to a valid input and that there is exactly one spent UTXO per input.

pallets/utxo/src/script.rs
iljakuklic: It has to do with `OP_CODESEPARATOR`. Bitcoin uses it in two ways, depending on script version.
1. The portion of the script starting at the last seen CODESEPATAOR is reinjected into the transaction input to be signed. This is before taproot.
2. The position of the last seen `OP_CODESEPARATOR` in message to be signed. This is used in Taproot script and is way simpler and more efficient.

The missing parameter gives you the script fragment from the last code separator onwards. We only implement the second approach so this is not needed. However, I like to keep the `chainscript` crate generic enough for it to be possible to implement the Bitcoin semantics on top of that too.

In general `OP_CODESEPARATOR` is not used very much, I did not find any decent examples with it, and value it brings I reckon is a bit questionable.

iljakuklic:
Rebased this again atop the token system. The conflicts were localized but a bit tricky to resolve, @sinitcin can you check I have not missed something?

PR #52, opened 2021-10-04
DO NOT MERGE.  This is for using `pallet-staking`'s logic to do staking. 

b-yap:
pallet-utxo
- `lib.rs`
  - added more config:
    - constants:
      - **`RewardReductionFraction`** - the percentage drop of a reward value. (e.g. 10%, 15%, etc.)
      - **`RewardReductionPeriod`** - the next block number to execute a drop of reward value.
      - **`MinimumStake`** - minimum value to start staking/validating
   - **`StakingHelper`** - a trait for performing staking OUTSIDE of pallet-utxo
 - added `Destinations`:
  - **`Stake`** - first time of the node to do staking. Requires the dummy `stash_account`, the `controller_account`, and the `rotate_keys` (for `pallet-session`). This is to satisfy the `pallet-staking`.
  - **`StakeExtra`** - if the validator wants to stake extra utxos (also `bond_extra` of `pallet-staking`). Can be done ONLY for existing validators.
  - **`WithdrawStake`** - stop validating and withdraw the staked utxos (also called as "locked utxos"). This can be done ONLY if the `unlock_stake` is executed.
 - added more storage:
   - **`StartingPeriod`** - stores the 1st block number of every new period
   - **`MLTCoinsAvailable`** - the extra mlt coins used to reward block authors, etc.
   - **`BlockAuthorRewardAmount`** - the reward value
   - **`BlockAuthor`** - the current block author
   - **`LockedUtxos`** - the "staked" utxos of the validators. Cannot be withdrawn.
   - **`StakingCount`** - tracks the number of utxos a controller has locked.
 - more events:
   - **`BlockAuthorRewarded`** - shows how much a validator is rewarded for producing a block
   - **`StakeUnlocked`** - shows a validator quiting the role.
   - **`StakeWithdrawn`** - shows a node withdrawing its staked/locked utxos; in other words, moving the utxo back to `UtxoStore`.
 - added Errors:
   - **`NoStakingRecordFound`**,
   - **`StakingAlreadyExists`** - a stash account CANNOT become a controller. And vice-versa
   - **`InvalidOperation`**
 - added more fields in `GenesisConfig`:
   - **`locked_utxos`** - the initial utxos locked (the initial authorities/validators)
   - **`extra_mlt_coins`** - the coins used for rewarding, etc.
   - **`initial_reward_amount`** - the initial block author reward amount  

- `authorship.rs` - to help look for the block author 

- `rewards.rs` - handles rewarding block authors 

- `staking.rs` - handles the staking. Also contains the trait definition of `StakeHelper`. For implementation, go to `runtime/src/staking.rs`.
  - **`fn get_account_id()`** - get account id conversion of public key.
  - **`fn stake()`**- handles the staking outside of `pallet-utxo.`
  - **`fn pause()`** - stops the validator from validating. It's not effective immediately; there's a waiting period time. (based on the `pallet-staking`)
  - **`fn withdraw()`** - withdraw funds back to free balance. 

- `staking_test.rs` - test cases for testing.  

runtime
- `lib.rs`
  - added the ff. pallets:
    - **`pallet-staking`** - the pallet used for its logic on staking. This follows the NPoS Consensus, BUT the configuration is only making use of its "validator" role.
    - **`pallet-session`** - this is a required pallet of `pallet-staking`
    -**`pallet-authorship`** - to identify the block author  

- `staking.rs` - implementation of StakeHelper.
  - **`fn get_account_id()`** - converts h256 to sr25519 Public and into the AccountId.
  - **`fn stake()`**- consists of `pallet-staking`'s `bond()`, `pallet-session`'s `set_key()` and `pallet-staking`'s `validate()`
  - **`fn pause()`** - performs `pallet-staking`'s `chill()` and `unbond()`. This does not mean that the funds can be withdrawn immediately; it should wait for the era to end. See the `SessionsPerEra` configuration of `pallet_staking::Config`, and the `Period` defined in `ShouldEndSession` of `pallet_session::Config`.
  - **`fn withdraw()`** - calls the `withdraw_unbonded()` of `pallet-staking` 

TODO:
 - add the `bond_extra` of `pallet-staking` (inside `runtime/staking.rs` when staking extra funds/utxos.
 - add more tests about the reward.
 - move those numerals like `40_000 * 100_000_000 * 1_000` AND `1_000 * 100_000_000 * 400_000_000` somewhere else.
 - ask Ben the minimum amount to spend for "withdrawing".
 - change reward reduction period to "yearly". 
  - change  `extra_mlt_coins` value to `200_000_000`. (inside `chain_spec.rs`)
  - benchmark the `fn unlock_stake()` call for  actual weight computation.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -1,6 +1,7 @@
 use node_template_runtime::{
-    pallet_utxo, AccountId, AuraConfig, BalancesConfig, GenesisConfig, GrandpaConfig, PpConfig,
+    pallet_utxo, AccountId, BalancesConfig, GenesisConfig, PpConfig,
     Signature, SudoConfig, SystemConfig, UtxoConfig, WASM_BINARY,
+    SessionConfig, StakingConfig, StakerStatus
 };
 use sc_network::config::MultiaddrWithPeerId;
 use sc_service::ChainType;
@@ -33,9 +34,23 @@ where
     AccountPublic::from(get_from_seed::<TPublic>(seed)).into_account()
 }
 
-/// Generate an Aura authority key.
-pub fn authority_keys_from_seed(s: &str) -> (AuraId, GrandpaId) {
-    (get_from_seed::<AuraId>(s), get_from_seed::<GrandpaId>(s))
+/// Keys required by pallet-session
+pub struct AuthorityKeys {
+    /// also called as `controller` account Id. see `pallet-staking`
+    account_id: AccountId,
+    stash_account_id: AccountId,
+    aura_id: AuraId,
+    grandpa_id: GrandpaId
+}
+
+/// Generate AuthorityKeys given a seed string
+pub fn authority_keys_from_seed(seed: &str) -> AuthorityKeys {
+    AuthorityKeys {
+        account_id: get_account_id_from_seed::<sr25519::Public>(seed),
+        stash_account_id: get_account_id_from_seed::<sr25519::Public>(&format!("{}//stash",seed)),
+        aura_id: get_from_seed::<AuraId>(seed),
+        grandpa_id: get_from_seed::<GrandpaId>(seed)
+    }
 }
 
 /// Return a list of bootnodes
@@ -111,7 +126,7 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
             testnet_genesis(
                 wasm_binary,
                 // Initial PoA authorities
-                vec![authority_keys_from_seed("Alice"), authority_keys_from_seed("Bob")],
+                vec![authority_keys_from_seed("Alice")],
                 // Sudo account
                 get_account_id_from_seed::<sr25519::Public>("Alice"),
                 // Pre-funded accounts
@@ -152,23 +167,60 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
 /// Configure initial storage state for FRAME modules.
 fn testnet_genesis(
     wasm_binary: &[u8],
-    initial_authorities: Vec<(AuraId, GrandpaId)>,
+    initial_authorities: Vec<AuthorityKeys>,
     root_key: AccountId,
     endowed_accounts: Vec<AccountId>,
     endowed_utxos: Vec<sr25519::Public>,
     _enable_println: bool,
 ) -> GenesisConfig {
+    const ENDOWMENT: u128 = 1 << 60;
+    // minimum balance set in the runtime. check `lib.rs` of runtime module.
+    const STASH: u128 = 500;
+
     // only Alice contains 400 million coins.
-    let genesis = endowed_utxos
+    let (genesis, locked_genesis) = endowed_utxos
         .first()
         .map(|x| {
             // may need to create a const variable to represent 1_000 and 100_000_000
-            pallet_utxo::TransactionOutput::new_pubkey(
+            let alice_genesis = pallet_utxo::TransactionOutput::new_pubkey(
                 1_000 * 100_000_000 * 400_000_000 as pallet_utxo::Value,
                 H256::from_slice(x.as_slice()),
-            )
+            );
+
+            let alice_auth_keys = authority_keys_from_seed("Alice");
+
+            // since alice is the starting validator, make sure she has some utxos being locked away
+            let alice_locked = pallet_utxo::TransactionOutput::new_stake(
+                // this is the minimum stake amount
+                40_000 * 100_000_000 * 1_000,
+                alice_auth_keys.stash_account_id,
+                alice_auth_keys.account_id,
+                vec![]
+            );
+
+            (alice_genesis, alice_locked)
         })
         .unwrap();
+
+    // initial_authorities also mean the starting validators in the chain.
+    let stakers = initial_authorities.iter()
+        .map(|auth_keys| (
+            auth_keys.stash_account_id.clone(),
+            auth_keys.account_id.clone(),
+            STASH,
+            // the role is `validator`. See pallet-staking
+            StakerStatus::Validator))
+        .collect::<Vec<_>>();
+
+    // Where aura and grandpa are initialized.
+    let session_keys = initial_authorities.iter().map(|x| (
+        x.stash_account_id.clone(),
+        x.stash_account_id.clone(),
+        node_template_runtime::opaque::SessionKeys {
+            aura: x.aura_id.clone(),
+            grandpa: x.grandpa_id.clone()
+        }
+    )).collect::<Vec<_>>();
 
     GenesisConfig {
         system: SystemConfig {
@@ -178,24 +230,40 @@ fn testnet_genesis(
         },
         balances: BalancesConfig {
             // Configure endowed accounts with initial balance of 1 << 60.
-            balances: endowed_accounts.iter().cloned().map(|k| (k, 1 << 60)).collect(),
+            balances: endowed_accounts.iter().cloned().map(|k| (k, ENDOWMENT)).collect(),
         },
-        aura: AuraConfig {
-            authorities: initial_authorities.iter().map(|x| (x.0.clone())).collect(),
-        },
-        grandpa: GrandpaConfig {
-            authorities: initial_authorities.iter().map(|x| (x.1.clone(), 1)).collect(),
-        },
+        // This has been initialized from the session config
+        aura: Default::default(),
+        // This has been initialized from the session config
+        grandpa: Default::default(),
         sudo: SudoConfig {
             // Assign network admin rights.
             key: root_key,
         },
         utxo: UtxoConfig {
             genesis_utxos: vec![genesis],
+            // The # of validators set should also be the same here.
+            // Currently forcing only 1 INITIAL AUTHORITY (Alice), hence only 1 locked-genesis.
+            // This means only ALICE has staked her utxo.
+            locked_utxos: vec![locked_genesis],
+            extra_mlt_coins: 50_000 as pallet_utxo::Value,
+            initial_reward_amount: 100 as pallet_utxo::Value,
             _marker: Default::default(),
         },
         pp: PpConfig {
             _marker: Default::default(),
         },
-    }
+        session: SessionConfig {
+            keys: session_keys
+        },
+        staking: StakingConfig {
+            validator_count: initial_authorities.len() as u32 + 5u32, // provide 5 more slots
+            // The # ofvalidators set should be the same number of locked_utxos specified in UtxoConfig.
+            minimum_validator_count: initial_authorities.len() as u32,
+            invulnerables: initial_authorities.iter().map(|x| x.account_id.clone()).collect(),
+            slash_reward_fraction: sp_runtime::Perbill::from_percent(10),
+            stakers,
+            .. Default::default()
+        }
+	}
 }
--- a/pallets/utxo/src/lib.rs
+++ b/pallets/utxo/src/lib.rs
@@ -1097,6 +1097,9 @@ pub mod pallet {
             spend::<T>(&signer, &tx)
         }
 
+        /// unlock the stake, if use wants to stop validating and wants to withdraw the locked utxos.
+        /// To withdraw, note that you need to check with the outside staking (a.k.a. pallet-staking)
+        /// the period of time (or era, as it is called) to when a withdrawal can be done.
         #[pallet::weight(T::WeightInfo::unlock_stake(1 as u32))]
         pub fn unlock_stake(
             origin: OriginFor<T>,
@@ -1114,7 +1117,9 @@ pub mod pallet {
     pub struct GenesisConfig<T: Config> {
         pub genesis_utxos: Vec<TransactionOutputFor<T>>,
         pub locked_utxos: Vec<TransactionOutputFor<T>>,
+        /// number of coins available for rewarding, esp. to block authors/producers.
         pub extra_mlt_coins:Value,
+        /// the amount to reward block authors/producers.
         pub initial_reward_amount:Value,
         pub _marker: PhantomData<T>,
     }
@@ -1146,7 +1151,6 @@ pub mod pallet {
                 if let Destination::Stake { stash_account:_, controller_account, rotate_keys:_ } =  &u.destination {
                     <StakingCount<T>>::insert(controller_account.clone(),1);
                 }
-                log::debug!("inserting {:?} to LockedUtxos:",u);
                 LockedUtxos::<T>::insert(BlakeTwo256::hash_of(&u), Some(u));
             });
         }
--- a/pallets/utxo/src/lib.rs
+++ b/pallets/utxo/src/lib.rs
@@ -19,6 +19,7 @@
 
 pub use header::*;
 pub use pallet::*;
+pub use authorship::*;
 
 #[cfg(test)]
 mod mock;
@@ -26,17 +27,26 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
+#[cfg(test)]
+mod staking_tests;
+
 #[cfg(feature = "runtime-benchmarks")]
 mod benchmarking;
 
+mod authorship;
 mod header;
+mod rewards;
 mod script;
+pub mod staking;
 pub mod weights;
 
 #[frame_support::pallet]
 pub mod pallet {
     use crate::TXOutputHeader;
     use crate::{OutputHeaderData, OutputHeaderHelper, TokenID};
+    use crate::{rewards::reward_block_author};
+    use crate::staking;
+    use crate::staking::{StakingHelper, validate_withdrawal};
     use chainscript::Script;
     use codec::{Decode, Encode};
     use core::convert::TryInto;
@@ -46,7 +56,8 @@ pub mod pallet {
         dispatch::{DispatchResultWithPostInfo, Vec},
         pallet_prelude::*,
         sp_io::crypto,
-        sp_runtime::traits::{BlakeTwo256, Dispatchable, Hash, SaturatedConversion},
+        sp_runtime::traits::{BlakeTwo256, Dispatchable, Hash, SaturatedConversion, Saturating},
+        sp_runtime::Percent,
         traits::IsSubType,
     };
     use frame_system::pallet_prelude::*;
@@ -59,7 +70,7 @@ pub mod pallet {
     use sp_core::{
         sp_std::collections::btree_map::BTreeMap,
         sp_std::{str, vec},
-        sr25519::{Public as SR25Pub, Signature as SR25Sig},
+        sr25519::Signature as SR25Sig,
         testing::SR25519,
         H256, H512,
     };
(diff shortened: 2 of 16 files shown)


node/src/chain_spec.rs
```diff
@@ -178,24 +230,40 @@ fn testnet_genesis(
...
             key: root_key,
         },
         utxo: UtxoConfig {
             genesis_utxos: vec![genesis],
+            // The # of validators set should also be the same here.
+            // Currently forcing only 1 INITIAL AUTHORITY (Alice), hence only 1 locked-genesis.
+            // This means only ALICE has staked her utxo.
+            locked_utxos: vec![locked_genesis],
+            extra_mlt_coins: 50_000 as pallet_utxo::Value,
+            initial_reward_amount: 100 as pallet_utxo::Value,
```
muursh, line 250: This is 100 base units rather than 100 MLT right?

muursh, line 250: What's the difference between initial reward and extra mlt?

runtime/src/lib.rs
```diff
@@ -297,13 +304,22 @@ impl pallet_template::Config for Runtime {
     type Event = Event;
 }
 
+parameter_types!{
+    pub const MinimumStake: u128 = 40_000 * 1_000 * 100_000_000;
+    pub const RewardReductionPeriod: BlockNumber = MINUTES * 5;
```
muursh, line 309: Does this mean we reduce the reward every 5 blocks?

node/src/chain_spec.rs
```diff
@@ -152,24 +167,61 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
...
 fn testnet_genesis(
     wasm_binary: &[u8],
-    initial_authorities: Vec<(AuraId, GrandpaId)>,
+    initial_authorities: Vec<AuthorityKeys>,
     root_key: AccountId,
     endowed_accounts: Vec<AccountId>,
     endowed_utxos: Vec<sr25519::Public>,
     _enable_println: bool,
 ) -> GenesisConfig {
+    const ENDOWMENT: u128 = 1 << 60;
```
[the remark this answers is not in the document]
b-yap: It's the initial balance of all `endowed_accounts`. It's to supply for the accounts in `pallet-balances`.

node/src/chain_spec.rs
b-yap: `initial_reward_amount` is the reward  for each block:

> "The first year each block includes 100 MLT tokens, the second year the reward is decreased to 75 MLT tokens and so on until 200 millions new MLT tokens are generated."

-------

The `extra_mlt_coins` is the `200_000_000 MLT` . That's part of my _todo_, 
> We have 200,000,000 to add in over time. Some of that will be from a faucet. Some of that could be just every block adding some mlt to the block reward

-------
for every block reward, it deducts from the `extra_mlt_coins`.

runtime/src/lib.rs
b-yap: yes

runtime/src/lib.rs
```diff
@@ -230,7 +237,7 @@ impl pallet_grandpa::Config for Runtime {
     type Event = Event;
     type Call = Call;
 
-    type KeyOwnerProofSystem = ();
+    type KeyOwnerProofSystem = Historical;
```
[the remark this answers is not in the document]
b-yap: To ensure that an account or a key is part of the validator set.
And the `Historical` is actually used when slashing balances of validators with bad rep; tracking down previous sessions.


I actually haven't tested this one yet; didn't find much examples to go by. We haven't yet discussed what to do when validators misbehave. I can set it back to `()`.

node/src/chain_spec.rs
```diff
@@ -152,24 +167,61 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
...
-            pallet_utxo::TransactionOutput::new_pubkey(
+            let alice_genesis = pallet_utxo::TransactionOutput::new_pubkey(
                 1_000 * 100_000_000 * 400_000_000 as pallet_utxo::Value,
                 H256::from_slice(x.as_slice()),
-            )
+            );
+
+            let alice_auth_keys = authority_keys_from_seed("Alice");
+
+            // since alice is the starting validator, make sure she has some utxos being locked away
```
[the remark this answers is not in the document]
b-yap: Can accounts be empty?

But they do need funds or UTXOs to perform a transaction, or to get a role. 
As a boot node, i think it needs money.
What would be the initial balance for each boot node?

node/src/chain_spec.rs
b-yap: yes. It's prolly **`100`** `* 100_000_000 * 1_000`...

runtime/src/staking.rs
```diff
@@ -0,0 +1,86 @@
...
+
+type StakeAccountId<T> =  <T as SysConfig>::AccountId;
+type LookupSourceOf<T> = <<T as SysConfig>::Lookup as StaticLookup>::Source;
+
+pub struct StakeOps<T>(sp_core::sp_std::marker::PhantomData<T>);
+impl <T: pallet_staking::Config + pallet_utxo::Config + pallet_session::Config> StakingHelper<T::AccountId> for StakeOps<T>
+    where StakeAccountId<T>: From<Public> + EncodeLike<AccountId32>
+{
+    fn get_account_id(pub_key: &H256) -> StakeAccountId<T> {
+      Public::from_h256(pub_key.clone()).into()
```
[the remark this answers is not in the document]
b-yap: Yes. :( This is because in the `type AccountId` of `runtime/lib.rs`, 
This portion here
```
/// Some way of identifying an account on the chain. We intentionally make it equivalent
/// to the public key of our transaction signing scheme.
pub type AccountId = <<Signature as Verify>::Signer as IdentifyAccount>::AccountId;
```

The `AccountId` is actually **`AccountId32`**.  It points to `sp_runtime::primitives`:
```
impl traits::IdentifyAccount for MultiSigner {
	type AccountId = AccountId32;
	fn into_account(self) -> AccountId32 {
		match self {
			Self::Ed25519(who) => <[u8; 32]>::from(who).into(),
			Self::Sr25519(who) => <[u8; 32]>::from(who).into(),
			Self::Ecdsa(who) => sp_io::hashing::blake2_256(who.as_ref()).into(),
		}
	}
}
```

runtime/src/lib.rs
```diff
@@ -370,6 +389,108 @@ impl pallet_contracts::Config for Runtime {
...
+}
+
+impl onchain::Config for Runtime {
+    type AccountId = AccountId;
+    type BlockNumber = BlockNumber;
+    type Accuracy = Perbill;
+    type DataProvider = Staking;
+}
+
+// pallet_staking_reward_curve::build! {
```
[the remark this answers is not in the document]
b-yap: that is... to be deleted.

b-yap:
I guess I have more questions, @muursh 
1. I only assumed about extra staking. Are we allowing this? Or only 1 time staking?
2. How much is the fee for staking? And how much for unstaking?
3. What to do with misbehaviours/misdemeanors/offenses? If we're slashing funds, that means accessing their UTXOs...
but we don't have that hash, unless the user provides it.
4. What's an SK?

node/src/chain_spec.rs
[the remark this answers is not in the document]
muursh: I guess they could. Why not split the genesis block to several UTXOs? Set aside a sk/UTXO for each boot node?

node/src/chain_spec.rs
muursh: Gotcha. I understand now

muursh:
> I guess I have more questions, @muursh
> 
> 1. I only assumed about extra staking. Are we allowing this? Or only 1 time staking?
> 2. How much is the fee for staking? And how much for unstaking?
> 3. What to do with misbehaviours/misdemeanors/offenses? If we're slashing funds, that means accessing their UTXOs...
>    but we don't have that hash, unless the user provides it.
> 4. What's an SK?

1. Yep we should allow multiple stakes
2. Good question, let's use 1 MLT (1 * 1_000 etc...) for now and change it in a bit
3. I guess we ignore them for now and just leave them be. 
4. Secret key. The same as a private key but we use the term secret so we can write sk/pk for secret and public keys because we're lazy :slightly_smiling_face:

runtime/src/staking.rs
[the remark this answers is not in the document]
muursh: Can this not work with pk directly? AccountID is just pk anyway right? How much work would be involved?

runtime/src/lib.rs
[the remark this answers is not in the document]
b-yap: This to compute inflation of the rewards to validators who produced a block and their respective nominators. 
But yes, to be deleted.

pallets/utxo/src/lib.rs
```diff
@@ -595,6 +754,25 @@ pub mod pallet {
...
+                    ensure!(!<LockedUtxos<T>>::contains_key(hash), "output already exists");
+                }
+                Destination::Stake {..} => {
+                    ensure!(output.value >= T::MinimumStake::get(), "output value must be equal or more than the set minimum stake");
+                    let hash = tx.outpoint(output_index);
+                    ensure!(!<LockedUtxos<T>>::contains_key(hash), "output already exists");
+                    new_utxos.push(hash.as_fixed_bytes().to_vec());
+                }
+                Destination::WithdrawStake{ outpoints, pub_key } => {
+                    ensure!(output.value == outpoints.len() as Value , "output value must be equivalent to the number of utxos to withdraw.");
```
altonen, line 770: Can you elaborate this check here? Isn't `output.value` some amount MLT we've staked and that we want to withdraw now? Either that's not true or I don't understand why we're comparing that value to the number of `outpoints`? Why are these two different types of numbers being compared?

pallets/utxo/src/staking.rs
```diff
@@ -0,0 +1,151 @@
...
+/// # Arguments
+/// * `output` - the output containing the `Destination::WithdrawStake`.
+/// * `hash` - is the outpoint/key to save the new utxo
+pub(crate) fn withdraw<T: Config>(output:&TransactionOutput<T::AccountId>,hash:H256) -> DispatchResultWithPostInfo {
+    if let Destination::WithdrawStake { outpoints, pub_key } =  &output.destination {
+        let controller_account = T::StakingHelper::get_account_id(pub_key);
+        T::StakingHelper::withdraw(&controller_account)?;
+
+        let mut sum:Value = 0;
+        outpoints.iter().for_each(|hash| sum += <LockedUtxos<T>>::take(hash).unwrap().value);
```
altonen, line 67: Is `hash` always going to be in `LockedUtxos` or could the code possibly panic here?

pallets/utxo/src/staking_tests.rs
```diff
@@ -0,0 +1,310 @@
...
+            inputs: vec![
+                TransactionInput::new_empty(karl_genesis)
+            ],
+            outputs: vec![
+                // ALICE withdraws her stake.
+                TransactionOutput::new_withdraw_stake(1,vec![alice_locked_utxo],H256::from(karl_pub_key)),
+                TransactionOutput::new_pubkey(98,H256::from(karl_pub_key))
+            ]
+        };
+        let karl_sig = crypto::sr25519_sign(SR25519,&karl_pub_key, &tx.encode()).unwrap();
```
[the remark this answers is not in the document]
altonen: IIRC `multiple_keys_test_ext()` creates a key context for the tests so user doesn't have to do anything when running these.

runtime/src/staking.rs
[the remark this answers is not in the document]
b-yap: ah, yes, I'll change it from `From<Public>` to **`From<[u8; 32]>`**, so the function will just be:
```
    fn get_account_id(pub_key: &H256) -> StakeAccountId<T> {
        pub_key.0.into()
    }
```

pallets/utxo/src/lib.rs
b-yap: Although now I'm thinking:
@muursh  do we allow withdrawal of a portion of the stake, or should it always be **all** of it?

Because if it's only a portion, meaning i have to change how the **unlock_stake** works.

runtime/src/staking.rs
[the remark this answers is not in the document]
muursh: That sounds like a good plan

pallets/utxo/src/staking_tests.rs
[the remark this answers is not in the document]
muursh: Sorry I was thinking more generically than the tests.

pallets/utxo/src/lib.rs
muursh: For now let's only support an entire withdrawal, they can re-stake a smaller amount if they decide they want to

pallets/utxo/src/staking_tests.rs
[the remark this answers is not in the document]
muursh: I assume so, I don't see why not @b-yap ?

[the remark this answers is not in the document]
b-yap: yep. That would be better.

b-yap:
https://github.com/mintlayer/core/pull/52#issuecomment-934188305

> Good question, let's use 1 MLT (1 * 1_000 etc...) for now and change it in a bit

@muursh 
Is it okay to deduct the fee from the utxos that were staked? Like the `40_000` staked mlt?

runtime/src/staking.rs
```diff
@@ -0,0 +1,101 @@
+use crate::Perbill;
```
altonen, line 1: Copyright header missing

runtime/src/staking.rs
```diff
@@ -0,0 +1,101 @@
...
+            vec![]
+        )?;
+
+        let validator_prefs = pallet_staking::ValidatorPrefs {
+            commission: Perbill::from_percent(0),
+            ..Default::default()
+        };
+
+        // validate
+        StakingPallet::<T>::validate(
```
altonen, line 50: I think code in this file can be simplified to use the return value of these `StakingPallet` calls. Like here the semicolon and `Ok(().into())` could be removed because `validate` seems to return `DispatchResultWithPostInfo` as well

runtime/src/staking.rs
```diff
@@ -0,0 +1,101 @@
...
+            StakingPallet::<T>::bond_extra(
+                RawOrigin::Signed(stake_ledger.stash).into(),
+                value.into()
+            )?;
+
+        } else {
+            log::error!("check sync with pallet-staking.");
+            Err(pallet_utxo::Error::<T>::NoStakingRecordFound)?
+        }
+        Ok(().into())
```
altonen, line 71: I think this can be simplified as well into something like
```
if let Some(...) {
    return StakingPallet::<T>::bond_extra(...);
}

log::error!("check sync with pallet-staking.");
return Err(pallet_utxo::Error::<T>::NoStakingRecordFound);
```

runtime/src/staking.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+
+    fn pause(controller_account: &StakeAccountId<T>) -> DispatchResultWithPostInfo {
+        // stop validating / block producing
+        StakingPallet::<T>::chill(RawOrigin::Signed(controller_account.clone()).into())?;
+
+        // get the total balance to free up
+        if let Some(stake_ledger) = <StakingPallet<T>>::ledger(controller_account.clone()) {
```
altonen, line 80: this as well

runtime/src/staking.rs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        Ok(().into())
+    }
+
+
+    fn withdraw(controller_account: &StakeAccountId<T>) -> DispatchResultWithPostInfo {
+        StakingPallet::<T>::withdraw_unbonded(RawOrigin::Signed(controller_account.clone()).into(),0)?;
+
+        Ok(().into())
```
altonen, line 99: Is this needed?

runtime/Cargo.toml
```diff
@@ -144,6 +175,13 @@ git = 'https://github.com/paritytech/substrate.git'
 version = '4.0.0-dev'
 branch = "master"
 
+## This might be needed during voting.
```
altonen, line 178: Is this still unsure or can this code be removed?

pallets/utxo/src/lib.rs
```diff
@@ -894,11 +1055,45 @@ pub mod pallet {
...
+        /// and at field `era`.
+        #[pallet::weight(T::WeightInfo::withdraw_stake(outpoints.len() as u32))]
+        pub fn withdraw_stake(
+            origin: OriginFor<T>,
+            controller_pub_key: H256,
+            outpoints: Vec<H256>
+        ) -> DispatchResultWithPostInfo {
+            ensure_signed(origin)?;
+            staking::withdraw::<T>(controller_pub_key,outpoints)?;
+            Ok(().into())
```
altonen, line 1085: Can the return value of `staking::withdraw()` be used directly?

pallets/utxo/src/lib.rs
```diff
@@ -354,30 +426,79 @@ pub mod pallet {
...
+    /// the utxo here is transferred back to `UtxoStore`.
+    #[pallet::storage]
+    #[pallet::getter(fn locked_utxos)]
+    pub(super) type LockedUtxos<T: Config> =
+        StorageMap<_, Identity, H256, Option<TransactionOutputFor<T>>, ValueQuery>;
+
+    /// this is to count how many stakings done for that controller account.
+    #[pallet::storage]
+    #[pallet::getter(fn staking_count)]
+    pub(super) type StakingCount<T: Config> = StorageMap<_, Identity,H256,(u64, Value), ValueQuery>;
```
altonen, line 466: Btw, should this save `Option((u64, Value))` instead? I don't actually even know what happens if you call `StakingCount::get()` with a key that doesn't exist. Is garbage returned or does the code panic?

altonen (commented):
There were some unused variables and includes and the code is not formatted. I also notice git has been doing something really weird because I see Ben's changes in the diff.

Am I correct in assuming that only way to test this right now is through the Polkadot UI?

node/src/chain_spec.rs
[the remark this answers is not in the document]
b-yap: @muursh  I never did ask; how many utxos should we create for Alice?

pallets/utxo/src/staking.rs
```diff
@@ -0,0 +1,179 @@
+
```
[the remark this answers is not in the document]
b-yap: added the copyright

runtime/Cargo.toml
b-yap: removed

pallets/utxo/src/lib.rs
b-yap: changed

node/src/chain_spec.rs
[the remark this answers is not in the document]
muursh: We should generate new key pairs for these utxos. Let's split alice in to 5 for now each equal utxos for the genesis block. This will change in the near future but it should be trivial to do

muursh:
Have you tested signing in the UI or utxo_tester.py and verified it works otherwise?

b-yap:
@muursh 

> Have you tested signing in the UI or utxo_tester.py and verified it works otherwise?
you mean **_our_** UI? I've only tested using the polkadot.js, so far.

not yet the utxo-tester since I needed the `author_rotateKeys`. I'll have to update that,

b-yap:
closing this botched git rebase PR, and moving to https://github.com/mintlayer/core/pull/68.

PR #53, opened 2021-10-04
The basic version Nft

sinitcin:
- Added support of NFT, in this version it's like a standard token just with additional data. 
- Removed the old TxOutputHeader. Now, we can use Enum instead of u128


pallets/utxo/rpc/runtime-api/src/lib.rs
```diff
@@ -25,6 +26,8 @@ sp_api::decl_runtime_apis! {
         // At the moment we have some problems with use serde in RPC, we can serialize and deserialize
         // only simple types. This approach allow us to return Vec<(TokenId, TokenName)> instead of
         // pallet_utxo_tokens::TokenListData
-        fn tokens_list() -> Vec<(u64, Vec<u8>)>;
+        fn tokens_list() -> Vec<(H256, Vec<u8>)>;
```
TheQuantumPhysicist, line 29: I think this function will only be useful for debugging. In reality, no application in the world will require having the list of ALL tokens on a blockchain. Imagine having that on Ethereum...

pallets/utxo/rpc/runtime-api/src/lib.rs
```diff
@@ -25,6 +26,8 @@ sp_api::decl_runtime_apis! {
...
+        // Getting NFT data
+        fn nft_read(id: H256) -> Option<(/* Data url */ Vec<u8>, /* Data hash */ [u8; 32])>;
```
TheQuantumPhysicist, line 31: I think limiting the data to 32 bytes is a big mistake. We should have a serialization mechanism that determines the data size. The simplest, is the first N bytes (4 is good enough), encoded with little or big endian, determining the size of the data. Then it can be as long as the user wants. The bigger it's, the more expensive the transaction will be.

TheQuantumPhysicist, line 31: I think the hash shouldn't be a `[u8, 32]`, but some `impl` to something that provides a generic hash interface. It's just nicer. Not a big deal.

TheQuantumPhysicist: Also, there can be a data type that "represents" an NFT, which will have both of these. I'm not a big fan of using tuples on arguments as they can be confusing. Instead of having a tuple commented somewhere hidden (like this) in the code, perhaps it's a better idea to just make a type that represents an NFT. That type will have tall the impls that will deal with hashes, initialization, checks, etc.

pallets/utxo/rpc/runtime-api/src/lib.rs
muursh: I think the chance of this surviving in to the testnet once we're happy everything is working properly is very slim

pallets/utxo/tokens/src/lib.rs
```diff
@@ -5,24 +5,33 @@ use serde::{Deserialize, Serialize};
...
+        // We can add another fields like:
+        //      pub number_format: NumberFormat,
+        //      pub image: UUID,
+        //      pub transaction: XXX,
+    },
+    Nft {
+        id: H256,
+        data_hash: [u8; 32],
+        data_url: Vec<u8>,
+        creator_pubkey: [u8; 32],
```
muursh, line 27: There's no guarantee of a 32 byte pubkey

pallets/utxo/tokens/src/lib.rs
```diff
@@ -5,24 +5,33 @@ use serde::{Deserialize, Serialize};
...
-    pub ticker: Vec<u8>,
-    pub supply: u128,
-    // We can add another fields like:
-    //      pub number_format: NumberFormat,
-    //      pub image: UUID,
-    //      pub transaction: XXX,
+pub enum TokenInstance {
+    Normal {
+        id: H256,
+        name: Vec<u8>,
```
muursh, line 15: I wonder if it's worth stripping the name out of the tx to reduce the size and just let the UI side handle this?

pallets/utxo/rpc/src/lib.rs
```diff
@@ -93,4 +101,22 @@ where
...
+        &self,
+        at: Option<<Block as BlockT>::Hash>,
+        id: H256,
+    ) -> Result<Option<(/* Data url */ Vec<u8>, /* Data hash */ [u8; 32])>> {
+        let api = self.client.runtime_api();
+        let at = BlockId::hash(at.unwrap_or_else(||
+            // If the block hash is not supplied assume the best block.
+            self.client.info().best_hash));
+
+        let runtime_api_result = api.nft_read(&at, id);
```
TheQuantumPhysicist, line 115: Could you please explain what's going on after this line?

pallets/utxo/rpc/runtime-api/src/lib.rs
sinitcin: Sure

pallets/utxo/tokens/src/lib.rs
muursh: This might be the time to serialise all of this, then we don't need to worry about pk size or extra data fields for different token types etc...

pallets/utxo/rpc/src/lib.rs
sinitcin: It's just a standard mechanism to call the API, actually, it will call `pallets/utxo/src/lib.rs:1083`:

```rust
    pub fn nft_read(
        id: H256,
    ) -> Option<(
        /* Data url */ frame_support::inherent::Vec<u8>,
        /* Data hash */ [u8; 32],
    )> {
        match Nft::<T>::get(id)? {
            TokenInstance::Nft {
                data_hash,
                data_url,
                ..
            } => Some((data_url, data_hash)),
            _ => None,
        }
    }
}
```

pallets/utxo/tokens/src/lib.rs
sinitcin: What do you think @TheQuantumPhysicist ? I can remove all tokens names, tickers, URLs from transactions but how we shall mapping the blockchain data with these?

muursh: So you don't necessarily need to remove it all, but ticker and name seems redundant

pallets/utxo/tokens/src/lib.rs
TheQuantumPhysicist: Agreed. Same for hashes. Perhaps using some `impl` for an interface for such components could be a good idea.

pallets/utxo/tokens/src/lib.rs
TheQuantumPhysicist: I think all this information (name, ticker, etc) should be part of a unified interface that we extract from data. The only thing that should be identifying a token, in general, is its `id`. Everything else should be defined by the user and/or the interface they choose to implement in their NFT.

pallets/utxo/src/lib.rs
```diff
@@ -753,21 +807,90 @@ pub mod pallet {
...
+                TransactionOutput::new_nft(
+                    *instance.id(),
+                    data_hash,
+                    data_url,
+                    H256::from(creator_pubkey)
+                ),
+            ],
+        };
 
+        let sig = crypto::sr25519_sign(SR25519, &creator_pubkey, &tx.encode()).unwrap();
```
altonen, line 874: Does this work if you mint using Polkadot UI? It didn't work for me with `send_to_address()` but now I'm no longer sure should I remove that logic given that NFTs also depend on it. I was anyway going to discuss this with @muursh anyway but any thoughts regarding this?

muursh: This will work if the node has access to the sk, as in the sk is stored using substrate's shitty km. Well it should work if that's the case. Does it fail if you try and sign something using the Alice's data which should be in the node's km? If it does do you have anything you can throw at me error wise?

TheQuantumPhysicist: Another `unwrap()`... please stop using it... this is a disaster waiting to happen.

altonen: IIRC it panicked or something similar at least when executed using a pallet call, I guess because it didn't have an access to the sk. I wonder why it works here but not with a pallet call...

TheQuantumPhysicist: It may or may not work... my thesis is that unless the developer can **prove** that it'll always succeed, we should never see `unwrap()` in the code.

pallets/utxo/src/lib.rs
```diff
@@ -230,12 +230,53 @@ pub mod pallet {
...
+    #[derive(Clone, Encode, Decode, Eq, PartialEq, PartialOrd, Ord, RuntimeDebug, Hash)]
+    pub enum TxHeaderAndExtraData {
+        NormalTx {
+            // Normal token ID
+            id: TokenId,
+        },
+        ConfidentialTx, /* not implemented yet */
+        Nft {
+            id: TokenId,
+            data_hash: [u8; 32],
```
TheQuantumPhysicist, line 243: Why are we storing the data hash with the header? This is kind of redundant. Shouldn't the transaction hash already have this information? If you need the data hash, you can just calculate it.

pallets/utxo/src/lib.rs
```diff
@@ -230,12 +230,53 @@ pub mod pallet {
...
+            data_url: String,
```
TheQuantumPhysicist, line 244: Can you please explain this?

pallets/utxo/src/lib.rs
```diff
@@ -230,12 +230,53 @@ pub mod pallet {
...
+            creator_pkh: H256,
```
TheQuantumPhysicist, line 245: Depending on how the implementation will be done, this may also be superfluous. We're gonna keep track of issuance transactions vs token IDs, we can just get this from the issuance transactions. No need to for this to be here.

pallets/utxo/tokens/src/lib.rs
sinitcin: > I think all this information (name, ticker, etc) should be part of a unified interface that we extract from data. The only thing that should be identifying a token, in general, is its `id`. Everything else should be defined by the user and/or the interface they choose to implement in their NFT.

May you please give me some examples? For me it is not clear.

sinitcin: What means `defined by the user and/or the interface`?

It's like a third-party website where artists could paint their pictures and there stored all information about NFT but in our blockchain, we store just ID and hash of the picture, is it?

TheQuantumPhysicist: Looking at Ethereum, the components of a transaction are destination, nonce, gasLimit, gasPrice and data. That's all. We don't need nonce, because we have UTXOs. The gas thing doesn't exist on substrate because they use weights instead (we'll have a discussion about that). And then, everyyyyyyything else should go into data. Notice that "data" can go into smart contracts, but in our case, we want to create a semi smart contract system where the data of a transaction is interpreted from the data in it instead of executing a smart contract. You start with an NFT trait/interface (which has at least the data and version method in it), and then you create NFTv1 trait that implements it and adds the fields you wanna add. All should be serializable within data. This should also contain the serialization version. So that when you create NFTv2 (which also implements NFT trait), it should be ready to be read. The serialization/deserialization should be customized.

A more general format of interfaces would be something like Ethereum's, where an interface is created as a json object that specifies what functions should be extracted from the data and its signature, but this can only be made after we define a format for the data in general.

To be clear, I'm not saying here how you should implement it. But what I'm saying is that you should not have any fixed fields whatsoever within your most basic NFT struct. Only `data` and `id` should be present. Everything else should be extracted dynamically based on the version.

pallets/utxo/src/lib.rs
```diff
@@ -230,12 +230,53 @@ pub mod pallet {
...
+                Self::Nft { .. } => true,
+                _ => true,
+            }
+        }
+    }
+
     /// Output of a transaction
     #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
     #[derive(Clone, Encode, Decode, Eq, PartialEq, PartialOrd, Ord, RuntimeDebug, Hash)]
+    #[repr(C)]
```
TheQuantumPhysicist, line 276: As explained previously, we should never use `repr(C)` anywhere, because this is architecture dependent. Alternatively we should use some serialization codec like SCALE or json or whatever. If you create a transaction on x86_64, then you move it to Raspberry Pi, everything will fail if you use `repr(C)`.

pallets/utxo/src/lib.rs
sinitcin: It's a hash of remote digital data that not include in transactions. It's impossible to calculate it without sending in a transaction all digital data that is unnecessary and might be huge.

pallets/utxo/src/lib.rs
```diff
@@ -478,38 +519,54 @@ pub mod pallet {
...
+                if let TxHeaderAndExtraData::NormalTx { id } = &output.header {
+                    Some((*id, output.value))
+                } else {
+                    None
+                }
             })
             .collect();
 
         // Check for token creation
         let tokens_list = <TokenList<T>>::get();
```
TheQuantumPhysicist, line 551: We're returning a vector here, and I think this is very inefficient... at least a BTreeSet should be returned out of this to have log(N) complexity when searching for tokens at worst instead of O(N). Or alternatively a HashSet to make it as close to O(1) as possible. Searching in a vector linearly seems like the worst way to do this, and can even open the door to DOS attacks.

I'm assuming here there are no duplicates. If there are no duplicates, you should convert the return value to a set or hash set before searching in it like you do below. Otherwise, we should have a discussion on how to do this.

Even better, you should be able to search for a token in the storage directly... why retrieve all tokens and search in them?!

pallets/utxo/src/tests.rs
```diff
@@ -516,3 +516,53 @@ fn test_send_to_address() {
         ));
     })
 }
+
+#[test]
```
TheQuantumPhysicist, line 520: I feel like the tests are not enough at all... there has to be tests also for all the negative cases, not only positive cases. Also tests that using fuzzing to prove that a data model is correct are encouraged.

pallets/utxo/src/lib.rs
sinitcin: You are right, I just forgot to remove that after our call. Done.

pallets/utxo/tokens/src/lib.rs
sinitcin: To be honest, I don't see any reason for serialization this place instead of having just enum type over here. Deriving Encode and Decode above the struct have already made serialization for the whole structure. 

But, if it might help later then okay. Done

pallets/utxo/src/lib.rs
```diff
@@ -753,21 +807,90 @@ pub mod pallet {
             ));
         }
 
+        let sig = crypto::sr25519_sign(
+            SR25519,
+            &sp_core::sr25519::Public::from_h256(public),
+            &tx.encode(),
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
sinitcin: Sure, at the moment I've changed it to 
```rust
let sig = crypto::sr25519_sign(SR25519, &pubkey, &tx.encode())
    .ok_or(DispatchError::Token(sp_runtime::TokenError::CannotCreate))?;
```

But in other places that were not relating to my task, we have it too. I hope, in time we remove these at all.

Fixed.

pallets/utxo/src/lib.rs
sinitcin: As we told before I should it rename to `data` and remove restrictions about data size. I've done it.

pallets/utxo/src/lib.rs
sinitcin: Does it mean that we need only `id`, `data`, and `owner` fields?

sinitcin: Why did I add `data_url` here, I'd like to have in the future the possibility like in the ETH to change metadata or link to digital data: https://eips.ethereum.org/EIPS/eip-1155#metadata

pallets/utxo/src/lib.rs
sinitcin: Answered in next line

pallets/utxo/tokens/src/lib.rs
sinitcin: @TheQuantumPhysicist I think, Ben's question was - The `TokenInstance::Normal` doesn't relate to NFT it's standard tokens without `data` field. But for this kind of token, we save their names and tickers. Do we need to do that?

pallets/utxo/src/lib.rs
sinitcin: @TheQuantumPhysicist Sorry but your answer is completely not related to this code, probably you have selected the wrong line? 

As an input we have Tx.Outputs that is equal `pub(crate) outputs: Vec<TransactionOutput<AccountId>>,`.
I should calculate all MLT that we use as a fee for transactions. Fee calculated like MLT from all outputs minus MLT from all inputs. 

We anyway will be going through all the `outputs` never mind what data container we will use, BTreeSet \ BTreeMap \ etc all of these require mapping the whole Vec.

What we can do `fn validate_transaction`? Reduce the number of Vectors and probably refactor it but it doesn't relate to this PR. @muursh

pallets/utxo/src/lib.rs
sinitcin: @muursh @altonen Unfortunately, it took time to double-check, and I can't answer yet.

pallets/utxo/src/tests.rs
sinitcin: As far as I remember, Ethereum uses for NFT only 3 tests. We have a bunch of tests for simple transactions. If we make a wrong transaction it will break other tests. Also, Substrate won't allow you to send the wrong types to your `pallet::call`. I will think about what we can test more but all things come to these pieces of code that have already been covered.

pallets/utxo/rpc/runtime-api/src/lib.rs
sinitcin: Okay, there is a problem:

If we change a tuple to a struct in `pallets/utxo/RPC/src/lib.rs:35`

```rust 
#[rpc]
pub trait UtxoApi<BlockHash> {
...
    #[rpc(name = "tokens_list")]
    fn tokens_list(&self, at: Option<BlockHash>) -> Result<Vec<TokenInstance>>;

```

We had this error:
```bash 
error[E0277]: the trait bound `pallet_utxo_tokens::TokenInstance: Deserialize<'_>` is not satisfied
  --> pallets/utxo/rpc/src/lib.rs:28:1
   |
28 | #[rpc]
   | ^^^^^^ the trait `Deserialize<'_>` is not implemented for `pallet_utxo_tokens::TokenInstance`
   |
   = note: required because of the requirements on the impl of `for<'de> Deserialize<'de>` for `Vec<pallet_utxo_tokens::TokenInstance>`
   = note: required because of the requirements on the impl of `DeserializeOwned` for `Vec<pallet_utxo_tokens::TokenInstance>`
   = note: this error originates in the attribute macro `rpc` (in Nightly builds, run with -Z macro-backtrace for more info)

```
Obviously, I need to make an implementation: 
```
 impl of `for<'de> Deserialize<'de>` for `Vec<pallet_utxo_tokens::TokenInstance>`
```
But `jsonrpc_derive` has a `feature = ["std"]` but the UTXO panel doesn't, when I tried to add the std feature to the UTXO panel I had a bunch of errors.

Any ideas guys? @altonen @iljakuklic

sinitcin:
```bash
     Compiling pp-api v3.0.0 (/home/anton/mintlayer/core/traits/pp-api)
  error[E0432]: unresolved import `sp_core::to_substrate_wasm_fn_return_value`
    --> /home/anton/.cargo/git/checkouts/substrate-7e08433d4c370a21/6e15de9/primitives/api/src/lib.rs:80:9
     |
  80 | pub use sp_core::to_substrate_wasm_fn_return_value;
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `to_substrate_wasm_fn_return_value` in the root

  For more information about this error, try `rustc --explain E0432`.
```
Weird

pallets/utxo/tokens/src/lib.rs
TheQuantumPhysicist: Nope. Same thing. Whether they're NFT or not, everything should be inside the data except for the id. The serialization format of the data should be used to extract token information.

sinitcin: > Nope. Same thing. Whether they're NFT or not, everything should be inside the data except for the id. The serialization format of the data should be used to extract token information.

Good, why we can't use just Enum instead of the version and data fields? Anyway, we are deriving Encode and Decode from the SCALE

pallets/utxo/src/lib.rs
sinitcin: As @TheQuantumPhysicist mentioned in DM, this comment related to change the type of the `TokenList`. I'll do that

pallets/utxo/src/lib.rs
muursh: So the default should be the tx being signed in the wallet, the average person won't be running a node with their sk stored in it. Signing using a locally stored sk is arguably an edge case. If the user tries it without a sk in km then that seems like a them problem to me

pallets/utxo/rpc/runtime-api/src/lib.rs
altonen: I don't know what is the proper Substrate way of doing things but looking at Parity's implementation, they don't use `jsonrpc_derive` for pallets so that's a bad sign. I also tried using JSON for a pallet call but it didn't go so well because of the `std` dependency. Are you able to use some other encoding?

pallets/utxo/src/lib.rs
altonen: @muursh  so the idea is that wallet only calls the `spend` pallet function which a signed transaction and possibly some other function to fetch the UTXOs of a given public key so that it can construct a TX? And those are the only two functions that wallet uses to interact with the UTXO system?

pallets/utxo/src/tests.rs
TheQuantumPhysicist: If I understand correctly, we're writing the token protocol from scratch. There has to be tests for higher and lower amount and equal amount of what exists. Tests for combining inputs and splitting outputs. There has to be tests for wrong token IDs. For too big data, too small data, invalid formats, etc. There's lots to be tested. This test is not enough, not by a long shot.

pallets/utxo/rpc/runtime-api/src/lib.rs
sinitcin: @altonen According to advice @TheQuantumPhysicist and Ben's decision we will remove RPC calls.

pallets/utxo/rpc/runtime-api/src/lib.rs
sinitcin: Removed

PR #54, opened 2021-10-06
Bech32 improvements

altonen:
This PR changes the way Bech32 is decoded. Instead of expecting it to decode to a valid scriptPubkey, decode the payload of Bech32 to a SCALE-encoding of `Destination` instead. This allows `send_to_address()` to use a more diverse set of `Destinations` when it's transferring the funds.


pallets/utxo/src/tests.rs
```diff
@@ -443,52 +443,12 @@ fn attack_double_spend_by_tweaking_input() {
...
-}
-
 #[test]
 fn test_send_to_address() {
     let (mut test_ext, alice_pub_key, _karl_pub_key) = new_test_ext_and_keys();
     test_ext.execute_with(|| {
-        // `addr` is bech32-encoded hash160(karl_pub_key)
-        let addr = "bc1q7pyaw92rh34mj6flsh7acccd7ayn4wf787ws4m";
+        // `addr` is bech32-encoded, SCALE-encoded `Destination::Pubkey(alice_pub_key)`
+        let addr = "bc1qrft7juyfhl06emj4zzrue5ljs6q39n2jalr4c40rhtcur647n0kwu2nvwe";
```
muursh, line 451: Shall we change the hrp from bc to ml? It would avoid confusion

muursh (commented):
All in all no real issues. @iljakuklic ?

pallets/utxo/src/tests.rs
altonen: ml for mainnet and tm for testnet?

muursh: Aye. Sounds good to me

pallets/utxo/src/lib.rs
```diff
@@ -841,10 +820,31 @@ pub mod pallet {
...
+            })?;
+
+            let mut tmp: &[u8] = &Vec::<u8>::from_base32(&data)
+                .or_else(|_| Err(DispatchError::Other("Base32 conversion failed")))?;
+
+            let dest: Destination<T::AccountId> = Destination::decode(&mut tmp).or_else(|_| {
+                Err(DispatchError::Other(
+                    "Failed to decode buffer into `Destination`",
+                ))
+            })?;
```
iljakuklic, line 846: Minor possible improvement (haven't tried, might be wrong):

Given `something: Result<_, _>`, we can replace the following pattern:
```rust
soething.or_else(|_| Err(foo))
```
with
```rust
something.map_err(|_| foo)
```

pallets/utxo/src/lib.rs
```diff
@@ -856,38 +856,33 @@ pub mod pallet {
...
-                    return Err("Invalid Bech32 address".into());
+            let tx_out = match dest {
+                Destination::Pubkey(pubkey) => {
+                    TransactionOutput::new_pubkey(value, H256::from(pubkey))
+                }
+                Destination::ScriptHash(hash) => TransactionOutput::new_script_hash(value, hash),
+                Destination::CreatePP(_, _) | Destination::CallPP(_, _) => {
+                    return Err(
+                        "OP_CREATE/OP_CALL UTXOs cannot be used with `send_to_address`".into(),
+                    )
```
iljakuklic, line 872: Wondering if we could use `dest` directly in the transaction output without this check. Would there be much harm in allowing `CreatePP` and `CallPP` and whatever other spend types we come up in the future to be encoded in a bech32 address?

pallets/utxo/src/lib.rs
```diff
@@ -841,10 +820,31 @@ pub mod pallet {
...
+                    Err(DispatchError::Other("Failed to decode address: mixed case"))
+                }
+                bech32::Error::InvalidChecksum => Err(DispatchError::Other(
+                    "Failed to decode address: invalid checksum",
+                )),
+                _ => Err(DispatchError::Other("Failed to decode address")),
+            })?;
+
+            let mut tmp: &[u8] = &Vec::<u8>::from_base32(&data)
+                .or_else(|_| Err(DispatchError::Other("Base32 conversion failed")))?;
```
iljakuklic, line 840: It feels to me that the users of the bech32 library should not have to deal with getting data out in two steps. I'd expect there to be a single higher-level call that goes from address directly to `Vec<u8>` data and the base32 conversion would be just an implementation detail of the bech32 libaray.

Also should HRP be checked here to be `ml` or `tm` depending on whether the node is running on testnet or mainnet and rejected otherwise?

iljakuklic: Sure I don't insist on the HRP checking being done in this PR but I think it should be checked eventually.

pallets/utxo/src/lib.rs
```diff
@@ -841,10 +820,29 @@ pub mod pallet {
...
+                    DispatchError::Other("Failed to decode address: invalid checksum")
+                }
+                bech32::Error::InvalidHrp => {
+                    DispatchError::Other("Failed to decode address: invalid HRP")
+                }
+                _ => DispatchError::Other("Failed to decode address"),
+            })?;
 
+            let mut slice: &[u8] = &data;
+            let dest: Destination<T::AccountId> = Destination::decode(&mut slice)
```
altonen, line 843: This is kind of ugly, does anyone know if there's a simpler way of converting `Vec<u8>` to a `&[u8]` without the compiler complaining about size not being known at compile time

sinitcin: `decode fn` requires implementation of `Input` trait for those types that might be decoded. Unfortunately, it does not implement for each case.

pallets/utxo/src/lib.rs
```diff
@@ -856,38 +853,32 @@ pub mod pallet {
...
-                    return Err("Invalid Bech32 address".into());
+            let tx_out = match dest {
+                Destination::Pubkey(pubkey) => {
+                    TransactionOutput::new_pubkey(value, H256::from(pubkey))
                 }
+                Destination::ScriptHash(hash) => TransactionOutput::new_script_hash(value, hash),
+                Destination::CreatePP(code, data) => {
+                    TransactionOutput::new_create_pp(value, code, data)
+                }
+                Destination::CallPP(acc, data) => TransactionOutput::new_call_pp(value, acc, data),
```
iljakuklic, line 869: I think this could be further simplified, just create the TransactionOutput directly (or create a new constructor function to do that):
```rust
TransactionOutput {
  destination: dest, value, header: /* whatever default header is? */
}
```

For a future PR, we should also think about extending these constructors to take output headers directly to make it harder to accidentally mix various token types.

zorvan (approved):

pallets/utxo/src/lib.rs
altonen: I don't understand why I wrote it like that, fixed

PR #57, opened 2021-10-12
utxo: Update the UI example to match recent developments

iljakuklic:
There are two relevant changes:
1. Had to make some last-minute adjustments to the signature PR prompted by the address handling
2. TXOutputHeader changed type


pallets/utxo/README.md
```diff
@@ -15,8 +15,7 @@ To run the test cases, just run command `cargo test`.
     "_enum": {
       "Pubkey": "Pubkey",
       "CreatePP": "DestinationCreatePP",
-      "CallPP": "DestinationCallPP",
-      "ScriptHash": "H256"
```
altonen, line 19: `ScriptHash` is still present in [code](https://github.com/mintlayer/core/blob/staging/pallets/utxo/src/lib.rs#L211), is this change correct?

PR #63, opened 2021-10-14
Fix functional tests to work with latest staging

TheQuantumPhysicist:
And maybe also add CI changes if review takes a long time


altonen (approved):
LGTM

PR #61, opened 2021-10-13
DO NOT MERGE. This is for using pallet-staking's logic to do staking. (part 2)

b-yap:
continuation from https://github.com/mintlayer/core/pull/52:

pallet-utxo
- `lib.rs`
  - added more config:
    - constants:
      - **`RewardReductionFraction`** - the percentage drop of a reward value. (e.g. 10%, 15%, etc.)
      - **`RewardReductionPeriod`** - the next block number to execute a drop of reward value.
      - **`MinimumStake`** - minimum value to start staking/validating
      - ****`StakeWithdrawalFee`** - the cost of withdrawing an unlocked staked utxo.**
   - **`StakingHelper`** - a trait for performing staking OUTSIDE of pallet-utxo
 - added `Destinations`:
  - **`Stake`** - first time of the node to do staking. Requires the dummy ~~`stash_account`, the `controller_account`,~~ **`stash pub key` and `controller pub key`** and the ~~`rotate_keys`~~ **`session key`** (for `pallet-session`). This is to satisfy the `pallet-staking`.
  - **`StakeExtra`** - if the validator wants to stake extra utxos (also `bond_extra` of `pallet-staking`). Can be done ONLY for existing validators.
  ~~- **`WithdrawStake`** - stop validating and withdraw the staked utxos (also called as "locked utxos"). This can be done ONLY if the `unlock_stake` is executed.~~
 - added more storage:
   - **`StartingPeriod`** - stores the 1st block number of every new period
   - **`MLTCoinsAvailable`** - the extra mlt coins used to reward block authors, etc.
   - **`BlockAuthorRewardAmount`** - the reward value
   - **`BlockAuthor`** - the current block author
   - **`LockedUtxos`** - the "staked" utxos of the validators. Cannot be withdrawn.
   - **`StakingCount`** - tracks the number of utxos a controller has locked, **and the total amount locked.**
 - more events:
   - **`BlockAuthorRewarded`** - shows how much a validator is rewarded for producing a block
   - **`StakeUnlocked`** - shows a validator quiting the role.
   - **`StakeWithdrawn`** - shows a node withdrawing its staked/locked utxos; in other words, moving the utxo back to `UtxoStore`.
 - added Errors:
   - **`NoStakingRecordFound`**,
   - **`StakingAlreadyExists`** - a stash account CANNOT become a controller. And vice-versa
   - **`InvalidOperation`**
   - **`OutpointDoesNotExist`**
   - **`FailedSignedTransaction`**
[...]

PR #68, opened 2021-10-15
 DO NOT MERGE. This is for using pallet-staking's logic to do staking. (part 3)

b-yap:
continuation from https://github.com/mintlayer/core/pull/61:

- @muursh  [about the split of utxos originally owned only by Alice](https://github.com/mintlayer/core/pull/61#discussion_r728800456): 
_Wait is this the new genesis block with alice split to several utxos held by different sks?_
  - I don't have the different sks yet; confused to being Bob, Charlie, Dave, etc? or completely different? So i settled with only Alice for now.

- @muursh  [about the number of validators in a set](https://github.com/mintlayer/core/pull/61#discussion_r728801437):
_Why 5?_
  - i've put a `TODO` comment for it.
 
- @sinitcin [on epoch](https://github.com/mintlayer/core/pull/61#issuecomment-943660917): 
_Just a question. How long does take an epoch_?
  - you mean a period? in the `pallet-timestamp`?

I removed the pallet-calls _`stake_using_account_id`_ and _`stake_extra`_ since they don't work yet, and this PR's changes is getting longer and harder to follow/review.

-------

pallet-utxo
- `lib.rs`
  - added more config:
    - constants:
      - **`RewardReductionFraction`** - the percentage drop of a reward value. (e.g. 10%, 15%, etc.)
      - **`RewardReductionPeriod`** - the next block number to execute a drop of reward value.
      - **`MinimumStake`** - minimum value to start staking/validating
      - ****`StakeWithdrawalFee`** - the cost of withdrawing an unlocked staked utxo.**
   - **`StakingHelper`** - a trait for performing staking OUTSIDE of pallet-utxo
 - added `Destinations`:
  - **`Stake`** - first time of the node to do staking. Requires the dummy ~~`stash_account`, the `controller_account`,~~ **`stash pub key` and `controller pub key`** and the ~~`rotate_keys`~~ **`session key`** (for `pallet-session`). This is to satisfy the `pallet-staking`.
  - **`StakeExtra`** - if the validator wants to stake extra utxos (also `bond_extra` of `pallet-staking`). Can be done ONLY for existing validators.
  ~~- **`WithdrawStake`** - stop validating and withdraw the staked utxos (also called as "locked utxos"). This can be done ONLY if the `unlock_stake` is executed.~~
 - added more storage:
   - **`StartingPeriod`** - stores the 1st block number of every new period
   - **`MLTCoinsAvailable`** - the extra mlt coins used to reward block authors, etc.
   - **`BlockAuthorRewardAmount`** - the reward value
   - **`BlockAuthor`** - the current block author
   - **`LockedUtxos`** - the "staked" utxos of the validators. Cannot be withdrawn.
   - **`StakingCount`** - tracks the number of utxos a controller has locked, **and the total amount locked.**
 - more events:
   - **`BlockAuthorRewarded`** - shows how much a validator is rewarded for producing a block
   - **`StakeUnlocked`** - shows a validator quiting the role.
   - **`StakeWithdrawn`** - shows a node withdrawing its staked/locked utxos; in other words, moving the utxo back to `UtxoStore`.
 - added Errors:
   - **`NoStakingRecordFound`**,
   - **`StakingAlreadyExists`** - a stash account CANNOT become a controller. And vice-versa
   - **`InvalidOperation`**
   - **`OutpointDoesNotExist`**
   - **`FailedSignedTransaction`**
 - added more fields in `GenesisConfig`:
   - **`locked_utxos`** - the initial utxos locked (the initial authorities/validators)
   - **`extra_mlt_coins`** - the coins used for rewarding, etc.
   - **`initial_reward_amount`** - the initial block author reward amount  
   - **changed the key of the `Utxostore` for the genesis utxos; added the index and a string _genesis_ on hashing.**

- `authorship.rs` - to help look for the block author 

- `rewards.rs` - handles rewarding block authors 

- **`mock.rs` - aura and the timestamp are not needed actually.**

- `staking.rs` - handles the staking. Also contains the trait definition of `StakeHelper`. For implementation, go to `runtime/src/staking.rs`.
  - **`fn get_account_id()`** - get account id conversion of public key.
  - **`fn stake()`**- handles the staking outside of `pallet-utxo.`
  - **`fn pause()`** - stops the validator from validating. It's not effective immediately; there's a waiting period time. (based on the `pallet-staking`)
  - **`fn withdraw()`** - withdraw funds back to free balance. 

- `staking_test.rs` - test cases for testing.  

runtime
- `lib.rs`
  - added the ff. pallets:
    - **`pallet-staking`** - the pallet used for its logic on staking. This follows the NPoS Consensus, BUT the configuration is only making use of its "validator" role.
    - **`pallet-session`** - this is a required pallet of `pallet-staking`
    -**`pallet-authorship`** - to identify the block author  

- `staking.rs` - implementation of StakeHelper.
  - **`fn get_account_id()`** - converts h256 to sr25519 Public and into the AccountId.
  - **`fn stake()`**- consists of `pallet-staking`'s `bond()`, `pallet-session`'s `set_key()` and `pallet-staking`'s `validate()`
  - **`fn pause()`** - performs `pallet-staking`'s `chill()` and `unbond()`. This does not mean that the funds can be withdrawn immediately; it should wait for the era to end. See the `SessionsPerEra` configuration of `pallet_staking::Config`, and the `Period` defined in `ShouldEndSession` of `pallet_session::Config`.
  - **`fn withdraw()`** - calls the `withdraw_unbonded()` of `pallet-staking` 

node
- `chainspec.rs`
  - create `AuthorityKeys` to represent the "validators", which will be used both by the `pallet-session` and the intial utxos of `pallet-utxo`.

TODO:
  - add the Keystore extension, to enable pallet calls of staking and staking extra.  and test it!!!
  -  complete the pallet calls staking and staking extra
  -  add the stake extra feature in utxo-tester
  - benchmark the `fn unlock_stake()` call for  actual weight computation.


pallets/utxo/src/lib.rs
```diff
@@ -351,30 +441,79 @@ pub mod pallet {
...
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     #[pallet::metadata(T::AccountId = "AccountId")]
     pub enum Event<T: Config> {
         TokenCreated(u64, T::AccountId),
         TransactionSuccess(TransactionFor<T>),
+
+        /// The block author has been rewarded with MLT Coins.
+        /// \[amount_paid, block_author_destination\]
+        //TODO: how to change this to TransactionOutput?
+        //TODO: currently the polkadot.js ui cannot translate Value and TransactionOutput
```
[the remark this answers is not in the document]
b-yap: updated/changed,

node/src/chain_spec.rs
```diff
@@ -152,23 +165,72 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
...
+       .fold( (vec![], vec![]), | (mut genesis_utxos, mut locked_utxos), auth_keys| {
+           let x_h256 = H256::from(auth_keys.controller.clone());
+           let x_stash_h256 = H256::from(auth_keys.stash.clone());
+
+           let num_of_utxos = 5;
+           let value = MLT_ORIG_SUPPLY / num_of_utxos;
+
+           for _ in 0..num_of_utxos {
+               //TODO the 5 split utxos are to be distributed by different public keys...
+               // but who are these public keys?
```
muursh, line 193: Us. We'll create them

pallets/utxo/src/mock.rs
```diff
@@ -16,29 +16,49 @@
...
+    pub lock_stash_map:BTreeMap<T::AccountId,T::AccountId>,
+    pub marker: PhantomData<T>
+}
+
+thread_local! {
+    pub static AUTHORITIES: RefCell<Vec<Public>> = RefCell::new(vec![]);
+    pub static MOCK_STAKING: RefCell<MockStaking<Test>> = RefCell::new(MockStaking::new());
+}
 
 pub const ALICE_PHRASE: &str =
```
muursh, line 63: We need to change alice. All her info is public

test/functional/feature_bob_stake_test.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""An example functional test
+
+Send a transaction from Alice to Bob, then spend Bob's transaction
```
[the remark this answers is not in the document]
muursh: Okay cool. I got confused for a few seconds, having not read the file name and couldn't work out how that comment related to the code

muursh (commented):
I think we could do with a few more functional tests too. Maybe several people staking and proving out staking "failing" when someone stakes <40_000MLT

pallets/utxo/src/lib.rs
```diff
@@ -351,30 +441,75 @@ pub mod pallet {
...
 
+    /// stores the first block number of every new period
+    #[pallet::storage]
+    #[pallet::getter(fn starting_period)]
+    pub(super) type StartingPeriod<T:Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
+
+    /// stores the extra MLTCoins for dispersing rewards
+    #[pallet::storage]
+    #[pallet::getter(fn coins_available)]
+    pub(super) type MLTCoinsAvailable<T> = StorageValue<_, Value, ValueQuery>;
```
TheQuantumPhysicist, line 452: Shouldn't there be predefined math that determines the reward based on the epoch/slot/height? It seems superfluous to have to store this.

pallets/utxo/src/lib.rs
```diff
@@ -351,30 +441,75 @@ pub mod pallet {
...
+    /// the utxo here is transferred back to `UtxoStore`.
+    #[pallet::storage]
+    #[pallet::getter(fn locked_utxos)]
+    pub(super) type LockedUtxos<T: Config> =
+    StorageMap<_, Identity, H256, Option<TransactionOutputFor<T>>, ValueQuery>;
+
+    /// this is to count how many stakings done for that controller account.
+    #[pallet::storage]
+    #[pallet::getter(fn staking_count)]
+    pub(super) type StakingCount<T: Config> = StorageMap<_, Identity,H256,Option<(u64, Value)>, ValueQuery>;
```
TheQuantumPhysicist, line 481: Minor: The formatting of this line is wrong. I believe we do have a setup rustfmt, right?

pallets/utxo/src/lib.rs
```diff
@@ -351,30 +441,75 @@ pub mod pallet {
...
+        /// Unlocked stake has been withdrawn.
+        /// \[total_stake, controller_account_H256_address\]
+        StakeWithdrawn(Value, H256)
     }
 
     #[pallet::hooks]
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
         fn on_finalize(block_num: T::BlockNumber) {
-            disperse_reward::<T>(&T::authorities(), block_num)
+            reward_block_author::<T>(block_num);
```
TheQuantumPhysicist, line 507: Would be great to explain what these do in a comment above them. Maybe I'll figure it out while reading.

pallets/utxo/src/lib.rs
```diff
@@ -386,7 +521,7 @@ pub mod pallet {
         trx.encode()
     }
 
-    fn disperse_reward<T: Config>(auths: &[H256], block_number: T::BlockNumber) {
+    fn reward_tx_validation<T: Config>(auths: &[H256], block_number: T::BlockNumber) {
```
TheQuantumPhysicist, line 524: Maybe a better name for this function would be `reward_block_validation`?
In fact, `disperse_reward` wasn't so bad. Also `grant_block_reward` sounds good.

pallets/utxo/src/lib.rs
```diff
@@ -401,7 +536,7 @@ pub mod pallet {
             .ok_or("Sub underflow")
             .unwrap();
```
TheQuantumPhysicist, line 537: Please avoid using unwrap. If this is an unrecoverable error and you want the program to crash, use expect() and explain in fair detail what happened and how our model is broken.

pallets/utxo/src/lib.rs
```diff
@@ -386,7 +521,7 @@ pub mod pallet {
         trx.encode()
     }
 
-    fn disperse_reward<T: Config>(auths: &[H256], block_number: T::BlockNumber) {
+    fn reward_tx_validation<T: Config>(auths: &[H256], block_number: T::BlockNumber) {
         let reward = <RewardTotal<T>>::take();
         let share_value: Value =
             reward.checked_div(auths.len() as Value).ok_or("No authorities").unwrap();
```
TheQuantumPhysicist, line 527: Besides the unwrap, where I explained how it should be treated in another comment.

I feel like this reward mechanism is very volatile. First, we're apparently "take()ing" the reward from the entry of the block. This means that, with no doubt, this block reward should be depleted completely by the end of this block. Using division like this potentially means we'll have remainder that will not be distributed. Besides, this will make the inflation schedule extremely chaotic. I don't see a reason for this. Simply give the rest of the reward to one of the authors. For simplicity, author number zero.

Second thing I have in mind, why do we have to use storage to determine the reward? The somehow glosses on the same question I had before. Why isn't the emission schedule clear and used?

pallets/utxo/src/lib.rs
```diff
@@ -401,7 +536,7 @@ pub mod pallet {
             .ok_or("Sub underflow")
             .unwrap();
 
-        log::debug!("disperse_reward:: reward total: {:?}", remainder);
+        log::debug!("reward_tx_validation:: reward total: {:?}", remainder);
         <RewardTotal<T>>::put(remainder as Value);
```
TheQuantumPhysicist, line 540: As per the explanation above, if I understand how this works, this should never be happen, and the remainder should *always* be zero.

pallets/utxo/src/lib.rs
```diff
@@ -640,6 +787,9 @@ pub mod pallet {
                         crate::script::verify(&tx, &input_utxos, index as u64, witness, lock)
                             .map_err(|_| "script verification failed")?;
                     }
+                    _ => {
```
TheQuantumPhysicist, line 790: Minor: Please do not use `_` at all unless we must. I understand you're doing this as placeholder, but preferably, add the arms that matches your enum and add the TODO notes there. We don't want to do the mistake of forgetting this and then failing to treat new types of outputs.

pallets/utxo/src/lib.rs
```diff
@@ -640,6 +787,9 @@ pub mod pallet {
...
+                        log::info!("TODO validate STAKE spending");
```
TheQuantumPhysicist, line 791: Quite frankly, I was looking forward to seeing how this is implemented to understand how all this would work... too bad.

pallets/utxo/src/lib.rs
TheQuantumPhysicist, line 790: Same comment about `_`

pallets/utxo/src/lib.rs
```diff
@@ -690,6 +840,31 @@ pub mod pallet {
...
+                Destination::StakeExtra(controller_pubkey) => {
+                    staking::stake_extra::<T>(controller_pubkey, output.value)?;
+
+                    let hash = tx.outpoint(index as u64);
+                    log::debug!("inserting to LockedUtxos {:?} as key {:?}", output, hash);
+
+                    let (mut num_of_utxos, mut total) = <StakingCount<T>>::get(&controller_pubkey).ok_or("cannot find the public key inside the stakingcount.")?;
+
+                    // update the total locked utxo values
+                    total = total.saturating_add(output.value);
```
TheQuantumPhysicist, line 860: Honestly, I'm not happy about the `saturating_add` being used everywhere here. If it ever saturates, it can break the invariants of the node and big problems can happen. We should never reach this point. So this (and all similar usages) should be changed to `checked_add` with an `expect()` on the error for the node to crash if this ever happens. This should provably never happen. Currently it's like hoping it won't happen.

pallets/utxo/src/lib.rs
```diff
@@ -903,11 +1078,46 @@ pub mod pallet {
...
+        /// unlock the stake, if user wants to stop validating and withdraw the locked utxos.
+        /// If used with `pallet-staking`, it uses the `BondingDuration`
+        /// to set the period/era on when to withdraw.
+        #[pallet::weight(T::WeightInfo::unlock_stake(1 as u32))]
+        pub fn unlock_stake(
+            origin: OriginFor<T>,
+            controller_pubkey: H256
+        ) -> DispatchResultWithPostInfo {
+            ensure_signed(origin)?;
+            staking::unlock::<T>(&controller_pubkey)?;
```
TheQuantumPhysicist, line 1091: I'm not sure, but shouldn't the stash owner be the only authority allowed to unlock a stake? What's the point of delegating a stake if a controller can control even spending it?

pallets/utxo/src/lib.rs
```diff
@@ -903,11 +1078,46 @@ pub mod pallet {
...
+        /// the field `unlocking` of datatype `UnlockChunk`,
+        /// and at field `era`.
+        #[pallet::weight(T::WeightInfo::withdraw_stake(outpoints.len() as u32))]
+        pub fn withdraw_stake(
+            origin: OriginFor<T>,
+            controller_pubkey: H256,
+            outpoints: Vec<H256>
+        ) -> DispatchResultWithPostInfo {
+            ensure_signed(origin)?;
+            staking::withdraw::<T>(controller_pubkey,outpoints)?;
```
TheQuantumPhysicist, line 1107: Same here about the controller overcontrol. If I understand correctly, a controller should only be able to stake. He shouldn't be allowed to spend the money (except for staking, which is a form of spending, depending on the protocol, which I'm not familiar with). So, spending otherwise (or withdrawal here) should only be authorized if the stash key signs the transaction to do this.

pallets/utxo/src/staking.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    /// stake more funds for the validator
+    fn stake_extra(controller_account:&AccountId, value:Value) -> DispatchResultWithPostInfo;
+
+    /// quitting the role of a validator.
+    fn pause(controller_account:&AccountId) -> DispatchResultWithPostInfo;
+
+    /// transfer balance from the locked state to the actual free balance.
+    fn withdraw(controller_account: &AccountId) -> DispatchResultWithPostInfo;
+
```
TheQuantumPhysicist, line 49: I kind of imagined the reward would go to the stash to be reused for staking. If one puts the stash as payee, will that reward be restaked?

runtime/src/staking.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn stake(stash_account: &StakeAccountId<T>, controller_account: &StakeAccountId<T>, session_key: &mut Vec<u8>, value: u128) -> DispatchResultWithPostInfo {
+        let controller_lookup: LookupSourceOf<T> = T::Lookup::unlookup(controller_account.clone());
+        let reward_destination = pallet_staking::RewardDestination::Staked;
+
+        // bond the funds
+        StakingPallet::<T>::bond(
+            RawOrigin::Signed(stash_account.clone()).into(),
+            controller_lookup,
+            value.into(),
+            reward_destination
```
TheQuantumPhysicist, line 49: I kind of imagined the reward would go to the stash to be reused for staking. If one puts the stash as payee, will that reward be restaked?

pallets/utxo/src/lib.rs
b-yap: yes, i will format everything once the review is done.

pallets/utxo/src/lib.rs
b-yap: i'll change them to `checked_add`.

runtime/src/staking.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct StakeOps<T>(sp_core::sp_std::marker::PhantomData<T>);
+impl <T: pallet_staking::Config + pallet_utxo::Config + pallet_session::Config> StakingHelper<T::AccountId> for StakeOps<T>
+    where StakeAccountId<T>: From<[u8; 32]>,
+          BalanceOf<T>: From<u128>
+{
+    fn get_account_id(pub_key: &H256) -> StakeAccountId<T> {
+        pub_key.0.into()
+    }
+
+    fn stake(stash_account: &StakeAccountId<T>, controller_account: &StakeAccountId<T>, session_key: &mut Vec<u8>, value: u128) -> DispatchResultWithPostInfo {
```
TheQuantumPhysicist, line 40: Why does the session_key have to be passed as bytes here? In the caller we do in the UTXO pallet, we convert this to Vec, then we make it mutable, and then we don't use neither the mutable nor the vec aspect of it. Maybe this is necessary for the runtime? I'm not sure. But from a pure rust this seems not to be optimal.

test/functional/feature_staking_less_than_minimum.py
```diff
@@ -0,0 +1,110 @@
...
+
+        alice = Keypair.create_from_uri('//Alice')
+        bob = Keypair.create_from_uri('//Bob')
+        bob_stash = Keypair.create_from_uri('//Bob//stash')
+
+        # fetch the genesis utxo from storage
+        utxos = list(client.utxos_for(alice))
+
+        # there's only 1 record of staking, which is alice.
+        assert( len(list(client.staking_count())) == 1 )
```
TheQuantumPhysicist, line 69: Please use assert_equal

runtime/src/staking.rs
TheQuantumPhysicist: This should probably be renamed to "lock coins to stake" or something

runtime/src/staking.rs
```diff
@@ -0,0 +1,113 @@
...
+            sesh_key,
+            vec![]
+        )?;
+
+        let validator_prefs = pallet_staking::ValidatorPrefs {
+            commission: Perbill::from_percent(0),
+            ..Default::default()
+        };
+
+        // validate
```
TheQuantumPhysicist, line 66: As discussed int he call, this is not validate, but a request to become a validator

pallets/utxo/src/lib.rs
TheQuantumPhysicist: As per the discussion in the call, we reward all the fees of a given block to the block-author.

pallets/utxo/src/lib.rs
```diff
@@ -104,6 +116,23 @@ pub mod pallet {
         Unapproved,
         /// The source account would not survive the transfer and it needs to stay alive.
         WouldDie,
+
+        /// When occurs during StakeExtra, use Destination::LockForStaking for first time staking.
+        /// When occurs during unstaking, it means there's no coordination with pallet-staking
+        NoStakingRecordFound,
+
+        /// Use Destination::StakeExtra for additional funds to stake.
+        StakingAlreadyExists,
```
TheQuantumPhysicist, line 125: `ControllerAccountAlreadyRegistered`? Or maybe `StakingAccountAlreadyRegistered`?

pallets/utxo/src/lib.rs
```diff
@@ -104,6 +116,23 @@ pub mod pallet {
         Unapproved,
         /// The source account would not survive the transfer and it needs to stay alive.
         WouldDie,
+
+        /// When occurs during StakeExtra, use Destination::LockForStaking for first time staking.
+        /// When occurs during unstaking, it means there's no coordination with pallet-staking
+        NoStakingRecordFound,
```
TheQuantumPhysicist, line 122: Maybe `ControllerAccountNotFound`?

pallets/utxo/src/staking.rs
```diff
@@ -0,0 +1,187 @@
...
+/// performing a withdrawal.
+/// # Arguments
+/// * `controller_pubkey` - the public key of the validator. In terms of pallet-staking, that's the
+/// controller, NOT the stash.
+/// * `outpoints` - a list of outpoints that were staked.
+pub(crate) fn withdraw<T: Config>(controller_pubkey: H256, outpoints: Vec<H256>) -> DispatchResultWithPostInfo {
+    validate_withdrawal::<T>(&controller_pubkey,&outpoints)?;
+
+    let controller_account = T::StakingHelper::get_account_id(&controller_pubkey);
+    T::StakingHelper::withdraw(&controller_account)?;
```
TheQuantumPhysicist, line 108: Please ensure the authentication of this process and make a test that attempts to withdraw without having authority to do so.

pallets/utxo/src/lib.rs
```diff
@@ -351,30 +441,75 @@ pub mod pallet {
...
     #[pallet::getter(fn reward_total)]
     pub(super) type RewardTotal<T> = StorageValue<_, Value, ValueQuery>;
 
+    #[pallet::storage]
+    #[pallet::getter(fn block_author_reward_amount)]
+    pub(super) type BlockAuthorRewardAmount<T> = StorageValue<_, Value, ValueQuery>;
+
+    #[pallet::storage]
+    #[pallet::getter(fn current_block_author)]
+    pub(super) type BlockAuthor<T> = StorageValue<_, Option<H256>, ValueQuery>;
```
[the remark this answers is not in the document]
iljakuklic: I had a quick look into this and it is actually not all that difficult. See PR https://github.com/mintlayer/core/pull/78

b-yap:
I am going to close this PR and open a new one, once I pulled from staging

PR #69, opened 2021-10-15
Add smart contract test

TheQuantumPhysicist:


test/functional/feature_smart_contract_test.py
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
```
altonen, line 2: I think some of the copyright belongs to RBB S.r.l

muursh:
Looks fine to me but @altonen and @zorvan are probably of more use

test/functional/feature_smart_contract_test.py
```diff
@@ -0,0 +1,144 @@
...
+        # with the instantiated contract
+        contractInstance = contract.ContractInstance(
+            ss58,
+            os.path.join(os.path.dirname(__file__), "metadata.json"),
+            substrate
+        )
+
+        # read the value of the flipper contract
+        result = contractInstance.read(alice, "get")
+        print('Current value of "get":', result.contract_result_data)
```
altonen, line 111: This could actually be `assert_equal(result.contract_result_data, 'False')"` (or however it's written) because the initial state of the flipper smart contract is `False` and the after the `OP_CALL` it could be verified to be `True`, proving that the call worked.

But I think I'll extend this code at some point to more comprehensively test the smart contract system so this can be left as is.

TheQuantumPhysicist: I'll fix this for now. Though, as you probably know, this is copied from smart contracts.

test/functional/feature_smart_contract_test.py
```diff
@@ -0,0 +1,144 @@
...
+                        code=os.path.join(os.path.dirname(__file__), "code.wasm"),
+                        data=[0xed, 0x4b, 0x9d, 0x1b],  # default() constructor selector
+                    )
+                ),
+            ]
+        ).sign(alice, [initial_utxo[1]])
+
+        # submit transaction and get the extrinsic and block hashes
+        (ext, blk) = client.submit(alice, tx0)
+
```
altonen, line 95: One thing we could also do at some point is to verify that the UTXO set contains correct UTXOs. Like here we could test that in addition to the UTXOs it had before anything was executed, we could check that it also contains UTXOs with values 0 and 50.

test/functional/feature_smart_contract_test.py
```diff
@@ -0,0 +1,144 @@
...
+                        dest_account=acc_id,
+                        input_data=bytes.fromhex(msg_data.to_hex()[2:]),
+                    )
+                ),
+            ]
+        ).sign(alice, [tx0.outputs[0]], [0])
+        (ext_hash, blk_hash) = client.submit(alice, tx1)
+
+        result = contractInstance.read(alice, "get")
+        print('Current value of "get":', result.contract_result_data)
```
altonen, line 140: This can be converted to assert

PR #73, opened 2021-10-18
Time locks

iljakuklic:
This implements absolute block-based time locks. In mimicks Bitcoin in separating script validation, which merely imposes restrictions on the `time_lock` field in the transaction, and the actual time lock checking.

### DIFF
--- a/pallets/utxo/src/lib.rs
+++ b/pallets/utxo/src/lib.rs
@@ -47,6 +47,7 @@ use utxo_api::UtxoApi;
 
 #[frame_support::pallet]
 pub mod pallet {
+    pub use crate::script::{BlockTime, RawBlockTime};
     use crate::sign::{self, Scheme};
     use crate::{OutputHeaderData, OutputHeaderHelper, TXOutputHeader, TokenID, TokenType};
     use bech32;
@@ -59,7 +60,7 @@ pub mod pallet {
         pallet_prelude::*,
         sp_io::crypto,
         sp_runtime::traits::{BlakeTwo256, Dispatchable, Hash, SaturatedConversion},
-        traits::IsSubType,
+        traits::{IsSubType, UnixTime},
     };
     use frame_system::pallet_prelude::*;
     use hex_literal::hex;
@@ -74,9 +75,7 @@ pub mod pallet {
         testing::SR25519,
         H256, H512,
     };
-    use sp_runtime::traits::{
-        AtLeast32Bit, Zero, /*, StaticLookup , AtLeast32BitUnsigned, Member, One */
-    };
+    use sp_runtime::traits::{AtLeast32Bit, Zero};
     use sp_runtime::DispatchErrorWithPostInfo;
 
     pub type Value = u128;
@@ -123,7 +122,7 @@ pub mod pallet {
 
     /// runtime configuration
     #[pallet::config]
-    pub trait Config: frame_system::Config {
+    pub trait Config: frame_system::Config + pallet_timestamp::Config {
         type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 
         type AssetId: Parameter + AtLeast32Bit + Default + Copy;
@@ -324,6 +323,7 @@ pub mod pallet {
     pub struct Transaction<AccountId> {
         pub(crate) inputs: Vec<TransactionInput>,
         pub(crate) outputs: Vec<TransactionOutput<AccountId>>,
+        pub(crate) time_lock: RawBlockTime,
     }
 
     impl<AccountId: Encode> Transaction<AccountId> {
@@ -349,6 +349,17 @@ pub mod pallet {
             self.inputs[index].witness =
                 crypto::sr25519_sign(SR25519, pk, &msg.encode())?.0.to_vec();
             Some(self)
+        }
+
+        pub fn check_time_lock<T: Config>(&self) -> bool {
+            match self.time_lock.time() {
+                BlockTime::Blocks(lock_block_num) => {
+                    <frame_system::Pallet<T>>::block_number() >= lock_block_num.into()
+                }
+                BlockTime::Timestamp(lock_time) => {
+                    <pallet_timestamp::Pallet<T> as UnixTime>::now() >= lock_time
+                }
+            }
         }
     }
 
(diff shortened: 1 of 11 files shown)


pallets/utxo/src/lib.rs
```diff
@@ -133,6 +133,50 @@ pub mod pallet {
...
+    /// Absolute time lock amount
+    #[derive(Eq, PartialEq, Clone, Copy, Encode, Decode, Debug, Default)]
+    #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+    pub struct AbsTimeLock(#[codec(compact)] u64);
+
+    impl AbsTimeLock {
+        // Values under this constant are considered to be a block number, time stamp otherwise.
+        // This constant is identical to what is used in Bitcoin.
+        // At 1min/block, the block numbers run out in about 900 years after the genesis block.
+        const THRESHOLD: u64 = 500_000_000u64;
```
iljakuklic, line 145: Here is one of the decision points: how to differentiate between real-time-based locks and block-height-based locks. There are many options.

* For now this is the same as Bitcoin (except it's `u64` rather than `u32`).
  * Value <500_000_000 is a block number, timestamp otherwise.
  * Block numbers are expected to run out in about 900 years with frequency of 1 block/min
  * The 500M timestamp corresponds to Nov 1985. Could be bumped to 1_000_000_000 which is Sep 2001. 
* Another option is to embrace Substrate types a bit more.
  * `Timestamp` in Substrate is a 64-bit quantity giving number of *milliseconds* since the start of Unix epoch
    * that gives 1000x more space for block numbers, threshold could be 1_000_000_000_000
    * has higher precision (probably useless)
* Another option is to make this into an enum with variants for `Timestamp` and `BlockHeight`.
  * This is arguably more extensible as more variants can be added (Bitcoin block height?).

This refers to the time_lock field in the transactions. There is a related issue how to encode the time lock quantity in the Script. It may but is not required to be the same as this. In particular, option (3) deviates from the way things are done in Bitcoin quite a bit and Script has stricter Bitcoin compatibility requirements than transaction format.

Also an extra possibility is to bypass the transaction field completely and only have time locks in script.

TheQuantumPhysicist: While it doesn't really matter, option 3 is what I would go for, but I'd like to ask first, would there be issues with scripting if we use option 3? Can you please provide an example of a possible problem that could happen if we choose 3? Because after all, if this will cause problems with scripting, we just go for 1. I'm not worried how much grand-grand-grand-grand-...-child will spend his MLT in 950 years.

iljakuklic: There is no fundamental issue with scripting and option 3. It does, however feel a bit more awkward in script. Currrently, all values in bitcoin script are interpreted as either raw bytes or as a signed integer in the sign-and-magnitude representation. This adds a new type which is an enum of two unsigned integers, so we'd have data on the stack interpreted in a new way for this operation only. Another issue I can see that if script wants to use arithmetic operations to get the lock time / block number, it will no longer be possible, at least not in a nice way.

It is possible to have a hybrid: Use an enum in the transaction and something closer to what Bitcoin does now in Script. When adding a new time lock type, we can add a new variant to the enum and extend the script by allocating a new opcode or something like that.

TheQuantumPhysicist: I think then option 1 is best, unless someone else has a different opinion. I don't see how bitcoin script can interpret an enum. Eventually, it'll just take a number, and the interpretation will be done in the script.

pallets/utxo/src/lib.rs
```diff
@@ -133,6 +133,50 @@ pub mod pallet {
         fn send_to_address(u: u32) -> Weight;
     }
 
+    /// Absolute time lock amount
+    #[derive(Eq, PartialEq, Clone, Copy, Encode, Decode, Debug, Default)]
+    #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+    pub struct AbsTimeLock(#[codec(compact)] u64);
+
+    impl AbsTimeLock {
+        // Values under this constant are considered to be a block number, time stamp otherwise.
```
TheQuantumPhysicist, line 142: I think this comment is incomplete

pallets/utxo/src/lib.rs
```diff
@@ -133,6 +133,50 @@ pub mod pallet {
...
+        // This constant is identical to what is used in Bitcoin.
+        // At 1min/block, the block numbers run out in about 900 years after the genesis block.
+        const THRESHOLD: u64 = 500_000_000u64;
+
+        pub fn from_u64(tl: u64) -> Self {
+            Self(tl)
+        }
+
+        pub fn from_blocks(num_blocks: u64) -> Option<Self> {
```
TheQuantumPhysicist, line 151: Perhaps a better way to implement this, instead of these "from_x" functions, make a getter that returns an enum, which has the type inside it based on the threshold. Something like `get_concrete` vs `get_raw`. Imagine how this will be used in code.

pallets/utxo/src/lib.rs
```diff
@@ -133,6 +133,50 @@ pub mod pallet {
...
+                cur_block >= self.0
+            } else {
+                cur_time >= self.0
+            }
+        }
+    }
+
+    impl PartialOrd for AbsTimeLock {
+        fn partial_cmp(&self, rhs: &Self) -> Option<core::cmp::Ordering> {
+            (self.is_blocks() == rhs.is_blocks()).then(|| self.0.cmp(&rhs.0))
```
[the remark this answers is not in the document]
iljakuklic: How is a "no lock" different from saying transaction starts being valid at block height 0?

pallets/utxo/src/lib.rs
TheQuantumPhysicist: Yep. This way, the underlying type for bitcoin scripts is always u64. But for the developer who wants to use this struct, it can be distinguished, if so they please.

pallets/utxo/src/lib.rs
iljakuklic: > There's possible no good reason to have any deviation.

There already is a subtle semantic deviation that is a consequence of block frequency. If you time lock something for certain number of blocks on Bitcoin vs on Mintlayer, then in terms of real time, Bitcoin will take about 10x longer to unlock.

PR #78, opened 2021-10-21
utxo: Use OptionQuery for UTXO store

iljakuklic:
This allows us to get rid of some Option wrapping and unwrapping.

### DIFF
--- a/pallets/utxo/src/lib.rs
+++ b/pallets/utxo/src/lib.rs
@@ -385,8 +385,7 @@ pub mod pallet {
 
     #[pallet::storage]
     #[pallet::getter(fn utxo_store)]
-    pub(super) type UtxoStore<T: Config> =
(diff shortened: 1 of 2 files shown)


pallets/utxo/src/lib.rs
```diff
@@ -386,7 +386,7 @@ pub mod pallet {
     #[pallet::storage]
     #[pallet::getter(fn utxo_store)]
     pub(super) type UtxoStore<T: Config> =
-        StorageMap<_, Identity, H256, Option<TransactionOutputFor<T>>, ValueQuery>;
+        StorageMap<_, Identity, H256, TransactionOutputFor<T>, OptionQuery>;
```
iljakuklic, line 389: Hmm now I see `OptionQuery` is the default for `StorageMap` so can get rid of the type param completely

b-yap: i see. well, i'll change what's mine, first,

PR #83, opened 2021-11-01
Staking and Rewards (part 4)

b-yap:
continuation of: https://github.com/mintlayer/core/pull/68

[Lukas' comment about changing public to account Id](https://github.com/mintlayer/core/pull/68#discussion_r729641012)
[offline pks](https://github.com/mintlayer/core/pull/68#discussion_r735269456)
[Use OptionQuery](https://github.com/mintlayer/core/pull/68#discussion_r733587976)
[Reward all to block author (no more `disperse_rewards`) ](https://github.com/mintlayer/core/pull/68#discussion_r732538279)
[DisableValidatorsThreshold set to 0](https://github.com/mintlayer/core/pull/68#discussion_r738953159)

test_keys.json - keys here are based on polkadot.js... for testing purpose. SUBJECT TO CHANGE. This needs to change using the actual keys.

note: ran cargo fmt on this.

TODO: more tests


pallets/utxo/src/authorship.rs
```diff
@@ -0,0 +1,55 @@
...
+    core::marker::PhantomData<(T, Inner, Validators)>,
+);
+
+/// The inner `FindAuthor` will point to the BlockAuthor's index position, given the digest.
+/// That index position points to an index of a list of validators, which this pallet doesn't have.
+/// The `Validators` is a list that can be used to retrieve the account Id.
+impl<T: Config, Inner: FindAuthor<u32>, Validators: ValidatorSet<T::AccountId>>
+    FindAuthor<Validators::ValidatorId> for FindAccountFromAuthorIndex<T, Inner, Validators>
+{
+    fn find_author<'a, I>(digests: I) -> Option<Validators::ValidatorId>
```
TheQuantumPhysicist, line 35: Can you please explain why we're implementing this ourselves? If I understand correctly, Aura has this already implemented. Why does a digest have to be involved? This sounds more like BABE way of doing things.

pallets/utxo/src/lib.rs
```diff
@@ -114,6 +125,25 @@ pub mod pallet {
...
+        InvalidOperation,
+
+        /// Problem signing a transaction with the public key
+        FailedSigningTransaction,
+
+        /// The hash outpoint key does not exist in the storage where it expects to be.
+        OutpointDoesNotExist,
+
+        /// The hash outpoint key already exists in the storage where it expects to be.
+        OutpointAlreadyExist,
```
TheQuantumPhysicist, line 146: `OutpointAlreadyExists`

pallets/utxo/src/authorship.rs
b-yap: I wanted to store the "block author" in the utxo side.

Yes, I'm utilizing Aura to identify the block author. I pass the digest to the **_`Inner`_** `FindAuthor`, which is Aura.


https://github.com/mintlayer/core/pull/83/files#diff-0ec06ea58bd455f09ce6b3bb4c2c1c0d37bda51c1e1be2151c560c9c973959ecR482

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,78 @@
...
+            };
+
+            if !<UtxoStore<T>>::contains_key(hash) {
+                <UtxoStore<T>>::insert(hash, utxo.clone());
+
+                <Pallet<T>>::deposit_event(Event::<T>::BlockAuthorRewarded(utxo));
+            }
+        } else {
+            log::warn!("no block author found for block number {:?}", block_number);
+            <RewardTotal<T>>::put(fees_total + reward_amount);
```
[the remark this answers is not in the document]
b-yap: i did...

pallets/utxo/src/lib.rs
```diff
@@ -220,6 +277,15 @@ pub mod pallet {
         CallPP(AccountId, bool, Vec<u8>),
         /// Pay to script hash
         ScriptHash(H256),
+        /// First attempt of staking.
+        /// Must assign a controller, in order to bond and validate. see pallet-staking
+        LockForStaking {
+            stash_account: AccountId, //TODO: change back to Public/H256 or something, after UI testing.
```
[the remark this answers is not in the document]
b-yap: I am removing those comments, based on what you told me about what Sergey said.

node/src/chain_spec.rs
```diff
@@ -23,19 +24,26 @@ pub fn get_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pair>::Pu
...
-    AccountPublic: From<<TPublic::Pair as Pair>::Public>,
-{
-    AccountPublic::from(get_from_seed::<TPublic>(seed)).into_account()
+/// Holds information about keys needed for the accounts
+#[derive(Default, Debug, Clone)]
+pub struct MltKeysInfo {
+    pub name: String,
+    pub sr25519_public_controller: sr25519::Public,
+    pub sr25519_public_stash: sr25519::Public,
+    pub ed25519_public: sp_core::ed25519::Public,
```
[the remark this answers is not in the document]
b-yap: ed25519 is for grandpa, block finalization.

node/src/command.rs
```diff
@@ -49,6 +59,74 @@ fn fetch_bootnode_list() -> Result<Vec<MultiaddrWithPeerId>, Box<dyn Error>> {
...
+            serde_json::from_str(&contents).map_err(|e| e.to_string())?;
+        for user in users["users"].as_array().ok_or("invalid json to extract user list")? {
+            let mut x: MltKeysFromFile =
+                serde_json::from_value(user.clone()).map_err(|e| e.to_string())?;
+            x.mlt_coins = x.mlt_coins * MLT_UNIT;
+            key_list.push(x.into());
+        }
+    } else {
+        log::debug!("failed to get keys from a file; using dummy values to populate keys");
+        //TODO: dummy values, in case the file doesn't exist.
```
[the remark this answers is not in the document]
b-yap: i removed it; placed in a file used ONLY for functional tests; and that would be the `--dev` mode.

runtime/src/lib.rs
```diff
@@ -128,11 +135,19 @@ pub const SLOT_DURATION: u64 = MILLISECS_PER_BLOCK;
...
+
+// number of slots available per era, of pallet-staking.
+pub const NUM_OF_VALIDATOR_SLOTS: u32 = 10;
 
 pub const MILLICENTS: Balance = 1_000_000_000;
 pub const CENTS: Balance = 1_000 * MILLICENTS;
 pub const DOLLARS: Balance = 100 * CENTS;
 
+/// The initial supply of mlt coins
+pub const MLT_ORIG_SUPPLY: Balance = 400_000_000_000 * MLT_UNIT;
```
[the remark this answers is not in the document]
b-yap: renamed it

b-yap:
@muursh , so i set the `--local` to use https://github.com/mintlayer/core/pull/83/files#diff-cdd1a61745cf1c21b3c5dcd75e3e521b994a2346d28e581d693b521ff9e5baf4

and for the functional tests to use the "already" known sks, for easier testing.

node/src/chain_spec.rs
```diff
@@ -53,10 +61,13 @@ fn get_bootnodes() -> Vec<MultiaddrWithPeerId> {
     ]
 }
 
-pub fn development_config() -> Result<ChainSpec, String> {
+pub fn development_config(endowed_accounts: Vec<MltKeysInfo>) -> Result<ChainSpec, String> {
     let wasm_binary = WASM_BINARY.ok_or_else(|| "Development wasm not available".to_string())?;
     let bootnodes = get_bootnodes();
 
+    // only Alice has sudo powers
```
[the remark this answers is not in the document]
b-yap: only for functional tests, because generating the keypair with seed `//Alice` is easier than the mnemonic.

node/src/chain_spec.rs
```diff
@@ -152,23 +143,77 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
...
 ) -> GenesisConfig {
-    // only Alice contains 400 million coins.
-    let genesis = endowed_utxos
-        .first()
-        .map(|x| {
-            // may need to create a const variable to represent 1_000 and 100_000_000
-            pallet_utxo::TransactionOutput::new_pubkey(
-                1_000 * 100_000_000 * 400_000_000 as pallet_utxo::Value,
-                H256::from_slice(x.as_slice()),
+    //TODO: clean up this code
```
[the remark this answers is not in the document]
b-yap: updated code; already getting from the files inside assets folder

node/src/chain_spec.rs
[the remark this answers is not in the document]
muursh: This is old Alice right?

node/src/command.rs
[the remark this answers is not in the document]
muursh: functional tests can run on a --release build

pallets/utxo/src/rewards.rs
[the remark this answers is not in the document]
muursh: Cool was just checking

muursh:
For testing (functional or otherwise) we can stick with the old SKs. Doesn't bother me. My only concern is that the new SKs don't go near github

pallets/utxo/src/lib.rs
```diff
@@ -699,6 +823,9 @@ pub mod pallet {
                         crate::script::verify(&tx, &input_utxos, index as u64, witness, lock)
                             .map_err(|_| "script verification failed")?;
                     }
+                    Destination::LockForStaking { .. } | Destination::LockExtraForStaking(_) => {
+                        log::info!("TODO validate STAKE spending");
```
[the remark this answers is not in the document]
b-yap: changed it to return an error, because staking utxos cannot be spent.

node/src/chain_spec.rs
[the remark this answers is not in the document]
b-yap: when using the `--dev` , it's the old Alice. `--local` is using the new Alice.

Although now we have `--release`, is that where the new Alice should be?

pallets/utxo/src/authorship.rs
b-yap: update: I've removed this file.

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+    } else {
+        log::warn!("problem adding the block author reward and the fees.");
```
TheQuantumPhysicist, line 99: Talking about the `get_reward_amount`, I think it should never fail... otherwise the node should panic. There has to be tests testing the full input space of this function to ensure it'll never fail. Which also means that it has to be simple enough. This is why I'm not happy with that loop in there... I seem to fail to understand the purpose of that loop.

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fn note_uncle(_author: T::AccountId, _age: T::BlockNumber) {
+        log::info!("TODO: no support for this. Or is there...?");
+    }
+}
+
+/// Returns the newly reduced reward amount for a Block Author.
+/// How much a reward is reduced, is based on the config's`RewardReductionFraction`.
+fn get_reward_amount<T: Config>(block_number: T::BlockNumber) -> u128 {
```
TheQuantumPhysicist, line 51: This function seems way more complex than what it should be doing. This function shouldn't include fees or anything else, which seems to be the case based on the implementation of the `reward_block_author` function. Please document what exactly is going on in here. I fail to understand how this works and fail to understand why there's a while loop (!) in side a reward calculation function.

muursh: Couldn't it be something like
` if blocknum < 100 {
    reward = 100; 
  } else if blocknum < 1000 {
    reward = 50; 
  } else {
    reward = 0;
  } `
  
  I mean obviously that's over simplified and messy but the idea holds...

muursh: @b-yap ?

muursh: 1 year of 1 min blocks is roughly 525600 blocks for example

muursh: ` if blocknum < 525600 { reward = 100; } else if blocknum < (525600 * 2) { reward = 75; }`

muursh: etc..

TheQuantumPhysicist: What Ben mentioned is legitimate. Another thing we can do is just have an array with the rewards, and pick the element based on the block number. Both ways are OK. The only issue with Ben's way is that we have to create multiple functions depending on the consensus rules we want (i.e., testnet, mainnet, etc). If all are the same, that works too. I'm OK with either way.

For me, the most important thing is to have a specific and fixed way to do this that's simple enough to be tested independently in unit tests.

Simplicity, simplicity, then more simplicity. Things that are complex can have mistakes.

muursh: I'm happy with the array solution too

muursh: This is different to the mainnet btw. In the mainnet it goes to 0, for the testnet it bottoms out at 1. By choice

TheQuantumPhysicist: The array solution will only work if both testnet and mainnet have the same structure. If they don't, we have to find some other solution and two different functions to be called, whether through a trait or something else.

Well, for the mean time, Carla, it seems that we both just hope to see something simple that's testable. Feel free to do whatever you want otherwise.

runtime/src/staking.rs
```diff
@@ -0,0 +1,161 @@
...
+
+        Ok(().into())
+    }
+
+    fn withdraw(stash_account: &StakeAccountId<T>) -> DispatchResultWithPostInfo {
+        // get the controller account, given the stash_account.
+        let controller_account = <StakingPallet<T>>::bonded(stash_account.clone())
+            .ok_or(pallet_utxo::Error::<T>::StashAccountNotFound)?;
+
+        let stake_ledger = <StakingPallet<T>>::ledger(controller_account.clone())
```
TheQuantumPhysicist, line 138: Please explain what's going on here. You're using the size of `unlocking` to figure out the current state. Please explain how you do that above the thing. Even better, move it to its own function and return an enum value representing your conclusion.

pallets/utxo/src/lib.rs
```diff
@@ -220,6 +279,20 @@ pub mod pallet {
         CallPP(AccountId, bool, Vec<u8>),
         /// Pay to script hash
         ScriptHash(H256),
+        /// First attempt of staking.
+        /// Must assign a controller, in order to bond and validate. see pallet-staking
+        LockForStaking {
+            stash_account: AccountId,
+            controller_account: AccountId,
+            session_key: Vec<u8>,
```
TheQuantumPhysicist, line 287: This is a very weird way of doing things. Why does the user have to provide a session key? Isn't that automatically generated from the controller key?

pallets/utxo/src/lib.rs
```diff
@@ -262,6 +335,48 @@ pub mod pallet {
...
+        /// * `stash_acount` - This is considered as the "bank", which holds the funds. The full meaning is found in `pallet-staking`.
+        /// * `controller_account` = can be considered as the "manager". The full meaning is found in `pallet-staking`.
+        /// * `session_key` - for every new validator candidate, you want to link it to a session key.
+        /// Generation of `session_key` is done through an rpc call `author_rotateKeys`.
+        /// See https://docs.substrate.io/v3/concepts/session-keys/
+        pub fn new_lock_for_staking(
+            value: Value,
+            stash_account: AccountId,
+            controller_account: AccountId,
+            session_key: Vec<u8>,
```
TheQuantumPhysicist, line 351: Again, same question... why does the user have to provide the session key instead of the software generating it from the controller key, as the docs of substrate say?
https://docs.substrate.io/v3/concepts/session-keys/#:~:text=Substrate%20provides%20the%20Session%20pallet,used%20for%20their%20intended%20purpose.

TheQuantumPhysicist:
Btw, I have a feeling that all your staking work can be moved to another pallet. No pressure about that now. Maybe after the release of testnet.

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fn note_uncle(_author: T::AccountId, _age: T::BlockNumber) {
+        log::info!("TODO: no support for this. Or is there...?");
+    }
+}
+
+/// Returns the newly reduced reward amount for a Block Author.
+/// How much a reward is reduced, is based on the config's`RewardReductionFraction`.
+fn get_reward_amount<T: Config>(block_number: T::BlockNumber) -> Value {
```
TheQuantumPhysicist, line 52: Please add unit tests in this file that calculate the reward for a few block numbers. Including 0, 1, 100, 1000, etc... including critical points where the reward amount changes before and during and after.

So if the block reward is supposed to change and block 12345, test 12344, 12345 and 123456, and compare with the expected amount.

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    1 // TODO: this is only testnet specific to reward at least 1
+}
+
+/// Rewards the block author with a utxo of value based on the `BlockAuthorRewardAmount`
+/// and the transaction fees.
+pub(crate) fn reward_block_author<T: Config>(block_number: T::BlockNumber) {
+    let fees_total = <RewardTotal<T>>::take();
+    if let Some(reward_amount) = get_reward_amount::<T>(block_number).checked_add(fees_total) {
```
TheQuantumPhysicist, line 79: These nested `if let`s make it hard to read the code (same applies to `get_reward_amount`). Usually the better way is to extract a value from within the `if let` (or match), then start another `if let` (or match).

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,102 @@
...
+        // give rewards only if a block author is found
+        if let Some(block_author) = <BlockAuthor<T>>::take() {
```
TheQuantumPhysicist, line 81: Again, please explain what's going on here. What does `take()` does. I know we're moving a block author from the DB. But why? Where did the author get into the DB come from?

test/functional/feature_staking_withdraw_not_validator.py
```diff
@@ -0,0 +1,83 @@
...
+        client.withdraw_stake(charlie,outpoints)
+
+        ledger = list(client.get_staking_ledger())
+        assert_equal(len(ledger[0][1]['unlocking']),0)
+        assert_equal(len(ledger[1][1]['unlocking']),0)
+
+        locked_utxos = list(client.utxos('LockedUtxos'))
+        # there should still be 2 utxo locked, no actual withdrawal happened.
+        assert_equal(len(locked_utxos),2)
+
```
TheQuantumPhysicist, line 80: Please also add a test to see if someone who owns the controller key can withdraw, and someone without a key can withdraw.

test/functional/test_framework/mintlayer/utxo.py
```diff
@@ -6,9 +6,10 @@
 import scalecodec
 import os
 import logging
+from _staking import Staking
 
 """ Client. A thin wrapper over SubstrateInterface """
-class Client:
+class Client(Staking):
```
TheQuantumPhysicist, line 12: Please don't use inheritance to add functionality... use composition.

test/functional/test_framework/mintlayer/_staking.py
```diff
@@ -0,0 +1,13 @@
+import substrateinterface
```
TheQuantumPhysicist, line 1: Please, no underscore before filenames.

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,207 @@
...
+
+/// Returns the newly reduced reward amount for a Block Author.
+/// How much a reward is reduced, will be based on the config's`RewardReductionFraction`.
+fn get_block_author_reward<T: Config>(block_number: T::BlockNumber) -> Value {
+    let reward_amount = T::InitialReward::get();
+    let reduction_period: T::BlockNumber = T::RewardReductionPeriod::get();
+    let reduction_fraction = T::RewardReductionFraction::get().deconstruct();
+
+    // no computation needed; current block number is still within the reduction period.
+    if block_number < reduction_period {
```
TheQuantumPhysicist, line 73: This is completely unnecessary. If your math below is correct, this result will be the natural consequence of it.

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+}
+
+
+/// checks at what period the given block number belongs to.
+fn get_block_period<T:Config>(block_number: T::BlockNumber) -> u8 {
+    let reduction_period: T::BlockNumber = T::RewardReductionPeriod::get();
+
+    // get at what period the current block number is at.
+    block_number.checked_div(&reduction_period)
```
TheQuantumPhysicist, line 57: I think this needs to be treated a little differently. When dividing integers, the only thing that can go wrong is if you divide by zero (or UINT_MIN/-1, which is not relevant here). Meaning: Your division here should never fail unless there's a node configuration issue. So, given that the reduction period is a constant, I suggest you `expect` that the result is correct, then continue. 

Returning `u8::MAX` here doesn't make any sense from a logical point of view.

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,207 @@
...
+    let reduction_period: T::BlockNumber = T::RewardReductionPeriod::get();
+    let reduction_fraction = T::RewardReductionFraction::get().deconstruct();
+
+    // no computation needed; current block number is still within the reduction period.
+    if block_number < reduction_period {
+        return reward_amount;
+    }
+
+    // The maximum period before approaching to 100% reduction
+    let max_period = (100u8 / reduction_fraction) - 1;
```
TheQuantumPhysicist, line 78: Are you sure the `- 1` is correct? In fact, `max_period` is ambiguous on what it means.

Let's define it in a better way. Call it, for example, `last_block_rewarded_period`. This way, we know that if the reduction is 25%, then:

1. First year it's 100%
2. Second year it's 75%
3. Third year it's 50%
4. Forth year it's 25%
5. Fifth year it's the default reward, which is 1% for testnet and 0% for mainnet. 

This way, your answer will be 100/25=4, so anything after 4th period is fixed. No `- 1` is needed.

muursh: So strictly the mainnet will start at 100MLT/block and reduce 25% a year until 200_000_000 MLT have been issued then 0 thereafter. The testnet is different in the sense it has no cap and never reaches 0. The next release of the testnet - the one that will break everything in this release will have the exact structure of the mainnet token reward schedule. We could almost leave the mainnet 0% out and just change this for the next release entirely - we're going to break everything with the outpoint changes and crypto updates anyway so backwards compatibility won't exist anyway.

With that said I agree the `-1` seems off.

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    reward_amount - updated_reduction_fraction.mul_ceil(reward_amount)
+}
+
+fn insert_to_utxo_store<T:Config>(block_number:T::BlockNumber, block_author:H256, reward:Value) {
+    let utxo = TransactionOutput::new_pubkey(reward, block_author.clone());
+
+    let hash = {
+        let b_num = block_number.saturated_into::<u64>();
+        BlakeTwo256::hash_of(&(&utxo, b_num, "author_reward"))
```
TheQuantumPhysicist, line 101: This is a very bad way of doing it, but let's let it pass for the testnet as I kind of see why you're doing it this way. Please, please oh, please make a note somewhere that this has to be changed. It seems extremely arbitrary.

There has to be a well-defined way to find the block reward, not an arbitrary emergent property from something that's being done to solve a technical issue.

Later, we'll change this to something like coinstake, where the previous locked stake is spent and is replaced with a UTXO that has the previous amount + reward.

pallets/utxo/src/lib.rs
```diff
@@ -379,26 +503,56 @@ pub mod pallet {
     #[pallet::getter(fn tokens_higher_id)]
     pub(super) type TokensHigherID<T> = StorageValue<_, TokenID, ValueQuery>;
 
+    //TODO: For rename, to differentiate from rewarding a block author.
     #[pallet::storage]
     #[pallet::getter(fn reward_total)]
     pub(super) type RewardTotal<T> = StorageValue<_, Value, ValueQuery>;
```
TheQuantumPhysicist, line 509: How hard is it to rename this to something like `TransactionFeesToReward`

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,207 @@
...
+    // As written on the definition of Take:
+    // Take a value from storage, removing it afterwards.
+    // This is taking a value of the RewardTotal storage, freeing it up.
+    let transaction_fees = <RewardTotal<T>>::take();
+
+    if let Some(reward_amount) = get_block_author_reward::<T>(block_number).checked_add(transaction_fees) {
+        // As written on the definition of Take:
+        // Take a value from storage, removing it afterwards.
+        // This is taking a value of the BlockAuthor storage, freeing it up.
+        match <BlockAuthor<T>>::take() {
```
TheQuantumPhysicist, line 124: Isn't this considered an unrecoverable failure? In what cases should we expect the author to not be found? This kind of stuff is what should be documented.

pallets/utxo/src/rewards.rs
TheQuantumPhysicist: Given that this should never fail, I think the right behavior is to just `panic!`. You should use `expect` to achieve that.

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,207 @@
...
+    } else {
+        //TODO: what's the actual behaviour (or if this happens at all)
+        log::warn!("problem adding the block author reward and the fees.");
+
+        <RewardTotal<T>>::put(transaction_fees);
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
TheQuantumPhysicist, line 143: `mod test`

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,207 @@
...
+    let max_period = (100u8 / reduction_fraction) - 1;
+
+    // current period of the given block number
+    let current_period = get_block_period::<T>(block_number);
+
+    // when current_period has reached or exceeded the maximum period of reduction,
+    // return the least amount to reward.
+    if current_period > max_period {
+        // TODO: as a note, this is only testnet specific to reward at least 1
+        return 1;
```
TheQuantumPhysicist, line 87: You can simply return this from `Config`. Something as `DefaultMinimumReward`

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,207 @@
...
+            assert_eq!(get_block_period::<Test>(53),10);
+        });
+
+    }
+
+    #[test]
+    fn get_block_author_reward_test() {
+        alice_test_ext().execute_with(|| {
+            // at Block 1, period is at 0, meaning no deduction of the reward.
+            assert_eq!(get_block_author_reward::<Test>(1), 100);
```
TheQuantumPhysicist, line 176: Great! I like the tests. Exactly what I was looking for. Good job!

pallets/utxo/src/rewards.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    if let Some(reward_amount) = get_block_author_reward::<T>(block_number).checked_add(transaction_fees) {
+        // As written on the definition of Take:
+        // Take a value from storage, removing it afterwards.
+        // This is taking a value of the BlockAuthor storage, freeing it up.
+        match <BlockAuthor<T>>::take() {
+            None => {
+                log::warn!("no block author found for block number {:?}", block_number);
+                // carry over the fees to the next block rewarding.
+                <RewardTotal<T>>::put(reward_amount);
```
TheQuantumPhysicist, line 128: The reason why this failure is bad and should never happen, is because we're moving the fees from one block to another... very weird way of doing things.

test/functional/feature_staking_unlock_and_withdraw.py
```diff
@@ -0,0 +1,110 @@
...
+        assert_equal(events[1].value['module_id'],'Staking')
+        assert_equal(events[1].value['event_id'], 'Unbonded')
+
+        assert_equal(events[2].value['module_id'],'Utxo')
+        assert_equal(events[2].value['event_id'], 'StakeUnlocked')
+
+        ledger = list(client.get_staking_ledger())
+        assert_equal(len(ledger),2)
+
+        time.sleep(500)
```
TheQuantumPhysicist, line 92: Usually sleeping to synchronize is a bad idea. Please explain why this is being used here. The better way is using queues that ensure ordered events. For example, if you know that the event you're waiting for, call it X happens specifically before event Y, then you can wait for Y.

Also polling the database consistently and wait for a change is also legitimate.

Sleeping to synchronize can make the tests fail depending on the computer it's used on.

test/functional/feature_staking_unlock_not_validator.py
```diff
@@ -0,0 +1,79 @@
...
+
+        client.unlock_request_for_withdrawal(charlie)
+
+        ledger = list(client.get_staking_ledger())
+        assert_equal(len(ledger[0][1]['unlocking']),0)
+        assert_equal(len(ledger[1][1]['unlocking']),0)
+
+        locked_utxos = list(client.utxos('LockedUtxos'))
+        # there should still be 2 utxos locked, no actual withdrawal happened.
+        assert_equal(len(locked_utxos),2)
```
TheQuantumPhysicist, line 76: Shouldn't we here wait a few blocks to ensure that the unlock doesn't change state?

TheQuantumPhysicist:
@b-yap  I see in the tests you're testing the logic of locking, unlocking and withdrawal. Can you write a test to test staking, eras, sessions, etc? A test that resembles how the system will work and stake? That will be the ultimate test we need.

node/src/chain_spec.rs
```diff
@@ -53,10 +53,56 @@ fn get_bootnodes() -> Vec<MultiaddrWithPeerId> {
...
+                // Pre-fund all accounts in the pallet-balance
+                endowed_accounts.clone(),
+            )
+        },
+        // Bootnodes
+        bootnodes,
+        // Telemetry
+        None,
+        // Protocol ID
+        None,
```
iljakuklic, line 91: WIll have to add protocol ID at some point to avoid collisions with other Substrate-based networks

b-yap:
continue with https://github.com/mintlayer/core/pull/93

PR #87, opened 2021-11-03
utxo: Add some tests for real-time time locks

iljakuklic:
These tests use `proptest` for randomized property-based testing.

By default, the `prop_*` functions are run with 256 different inputs.

### DIFF
--- a/pallets/utxo/src/script.rs
+++ b/pallets/utxo/src/script.rs
@@ -162,11 +162,17 @@ pub fn verify<AccountId: Encode>(
 }
 
 #[cfg(test)]
-mod test {
+pub(crate) mod test {
     use super::*;
     use chainscript::Context;
     use core::time::Duration;
     use sp_core::sr25519;
+    use proptest::prelude::*;
+
+    // Generate block time in seconds
+    pub fn gen_block_time_real() -> impl Strategy<Value = RawBlockTime> {
+        (RawBlockTime::THRESHOLD..3 * RawBlockTime::THRESHOLD).prop_map(RawBlockTime::new)
+    }
 
     #[test]
     fn test_parse_pubkey() {
--- a/pallets/utxo/src/tests.rs
+++ b/pallets/utxo/src/tests.rs
@@ -16,8 +16,8 @@
 // Author(s): C. Yap
 
 use crate::{
-    mock::*, BlockTime, Destination, RewardTotal, TokenList, Transaction, TransactionInput,
-    TransactionOutput, UtxoStore, Value,
+    mock::*, BlockTime, Destination, RawBlockTime, RewardTotal, TokenList, Transaction,
+    TransactionInput, TransactionOutput, UtxoStore, Value,
 };
 use chainscript::{opcodes::all as opc, Builder};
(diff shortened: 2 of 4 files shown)


pallets/utxo/src/tests.rs
```diff
@@ -601,3 +602,96 @@ fn test_send_to_address() {
         );
     })
 }
+
+// Generate block time in seconds
+fn gen_block_time_real() -> impl Strategy<Value = RawBlockTime> {
+    (RawBlockTime::THRESHOLD..RawBlockTime::THRESHOLD + 1_000_000_000).prop_map(RawBlockTime::new)
```
muursh, line 608: What is this 1_000_000_000?

muursh (approved):
Looks good to me. Just not sure where the 1bn comes from

pallets/utxo/src/tests.rs
iljakuklic: That is just an arbitrary upper bound for generated timestamps. Can change it to `2 * THRESHOLD`, that may be more clear.

iljakuklic: Just an arbitrary upper bound fr generated timestamps. The result is `3 * THRESHOLD` so can change it to that for more clarity.

muursh: Either that or just a comment. I'm fine with either.

pallets/utxo/src/script.rs
```diff
@@ -32,7 +32,7 @@ impl RawBlockTime {
     // Values less than this constant are considered to be a block number, other values are
     // interpreted as UNIX time stamp. This constant is identical to what is used in Bitcoin.
     // At 1min/block, the block numbers run out in about 900 years after the genesis block.
-    const THRESHOLD: u64 = 500_000_000u64;
+    pub const THRESHOLD: u64 = 500_000_000u64;
```
TheQuantumPhysicist, line 35: I think you don't have to make this public if you make the tests under the `#[tests]`. This is a huge problem in C++, that you cannot access private variables in tests. But I think rust solved that problem. Please double check.

iljakuklic: I was thinking of something like:

```rust
#[cfg(test)]
const THRESHOLD: u64 = 500_000_000u64;
#[cfg(test)]
pub const THRESHOLD: u64 = 500_000_000u64;
```

But found it a bit too verbose for a small issue like this. If Rust has a better mechanism, that's great.

iljakuklic: So in the end, I moved the test helper function that uses the THRESHOLD constant into the `test` submodule of this module so it has access to struct fields without making them public. The function is then imported in `test.rs`.

TheQuantumPhysicist (approved):

iljakuklic:
Added two more tests, one checking time locks for `(now - 1, now, now + 1)`, other to check overflow for timestamps close to `u64::MAX`.

muursh:
New changes look good to me too

TheQuantumPhysicist (approved):
Looks good to me!

PR #92, opened 2021-11-06
C2X transfer refunds

altonen:
This PR introduces a new destination type called `Destination::FundPP` which is used solely for funding a smart contract. It also implements refunding for C2X transfers as previously all excess funds were burned, now they're returned back to the contract.

This branch is built on top of `pp-validation` so that has to be approved and merged first. That's why this is a draft for now.


pallets/utxo/src/lib.rs
```diff
@@ -267,7 +267,9 @@ pub mod pallet {
         CreatePP(Vec<u8>, Vec<u8>),
         /// Pay to an existing contract. Takes a destination account,
         /// whether the call funds the contract, and input data.
```
royster57, line 269: I think the comment here needs to be updated `CallPP` no longer takes the boolean of whether the call funds the contract.

pallets/utxo/src/lib.rs
```diff
@@ -377,19 +379,22 @@ pub mod pallet {
...
-        pub fn new_call_pp(
-            value: Value,
-            dest_account: AccountId,
-            fund: bool,
-            input: Vec<u8>,
-        ) -> Self {
+        pub fn new_call_pp(value: Value, dest_account: AccountId, input: Vec<u8>) -> Self {
             Self {
                 value,
                 destination: Destination::CallPP(dest_account, fund, input),
```
royster57: How does this compile? The `fund` parameter is not defined and also is no longer part of the `CallPP` variant

altonen: This is an outdated diff and the build works with latest changes

pallets/utxo/src/lib.rs
```diff
@@ -1380,16 +1392,18 @@ impl<T: Config> crate::Pallet<T> {
...
 
     let mut outpoints = outpoints.clone();
     outpoints.sort();
 
     for outpoint in outpoints.iter() {
         let tx = <UtxoStore<T>>::get(&outpoint).ok_or("UTXO doesn't exist!")?;
         match tx.destination {
-            Destination::CallPP(_, _, _) => {
+            Destination::FundPP(_) => {
+                total_value += tx.value;
```
TheQuantumPhysicist, line 1406: `checked_add`?

PR #99, opened 2021-11-09
chain_spec: Protocol ID for testnet

iljakuklic:
This is used to differentiate between Substrate-based chains


altonen (commented):
LGTM but I think this conflicts with the latest PR of @b-yap

iljakuklic:
> LGTM but I think this conflicts with the latest PR of @b-yap

This builds on top of it, the target branch is set to `rewards_and_stakig`

b-yap (approved):

PR #100, opened 2021-11-09
rewards_and_staking: fix the issue of exceeding u64 value of transaction fee

b-yap:


iljakuklic:
First of all, congratulations for taking PR #100 :smiley:. Could you briefly elaborate on the issue? Did the node reject the test due to `u64` overflow or does the node need to be fixed to handle this? Coin/token amounts are usually represented as `u128` so it seems a bit weird.

TheQuantumPhysicist:
The difference between inputs and outputs, which is the reward for miners/stakers, calculated in substrate by default with a u64 type. So, if the difference is too high, I decided to make it fail because that'll cause burning coins

iljakuklic (approved):

TheQuantumPhysicist (approved):

PR #103, opened 2021-11-10
staking: added commands for staking, staking extra, unlock, withdraw

b-yap:
you can update the code, because there are some repeating logic.

However the commands are there, which was my goal.

### DIFF
--- a/scripts/wallet.py
+++ b/scripts/wallet.py
@@ -55,6 +55,103 @@ def keygen(args):
     print('Mnemonic   :', mnemonic)
     print_key_info(keypair)
 
+
+def unlock(args):
+    acct = Account(args)
+    acct.client.unlock_request_for_withdrawal(acct.keypair)
+
+def withdraw(args):
+    acct = Account(args)
+    acct.client.withdraw_stake(acct.keypair)
+
+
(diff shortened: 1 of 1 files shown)


scripts/wallet.py
```diff
@@ -55,6 +55,103 @@ def keygen(args):
...
+    for (h, u) in acct.mlt_utxos():
+        utxos.append((h, u))
+        utxo_value += u.value
+        if utxo_value >= amount:
+            break
+    if utxo_value < amount:
+        raise Exception('Not enough funds')
+
+    fee = int(MLT_UNIT / 10) # TODO
+    if fee >= 2 ** 64:
```
altonen, line 88: Does this check make sense? Isn't `MLT_UNIT` a constant

iljakuklic: `MLT_UNIT` is 1 MLT, MLT_UNIT / 10 is 0.1 MLT hardcoded fee. I have an update where you can specify the fee MLT amount on the command line for payments

iljakuklic (approved):
Great job. Thank you

TheQuantumPhysicist (approved):

PR #105, opened 2021-11-10
staking: adding the locked balance; add 2 `withdrawal era` sub-command

b-yap:
Added a `withdrawal_era` command so the user will know at what era he/she will be able to withdraw. Also indicates the current era.

In the balance, added the `Total Locked`.

note: feel free to clean the code.

### DIFF
--- a/scripts/wallet.py
+++ b/scripts/wallet.py
@@ -31,14 +31,32 @@ class Account:
     def utxos(self):
         return self.client.utxos_for(self.keypair)
 
+    def locked_mlt_utxos(self):
+        return self.client.locked_utxos_for(self.keypair)
+
     def mlt_utxos(self):
         return ((h, u) for (h, u) in self.utxos() if u.data is None)
+
+    def locked_mlt_balance(self):
+        res = sum(u.value for (_, u) in self.locked_mlt_utxos())
+        if res == 0:
+            return 0
+        else:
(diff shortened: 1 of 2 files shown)


scripts/wallet.py
```diff
@@ -55,11 +73,15 @@ def keygen(args):
     print('Mnemonic   :', mnemonic)
     print_key_info(keypair)
 
-
 def unlock(args):
     acct = Account(args)
     acct.client.unlock_request_for_withdrawal(acct.keypair)
 
+def withdrawal_era(args):
```
altonen, line 80: Is this just a debug function or did you mean to return something from this?

test/functional/test_framework/mintlayer/utxo.py
```diff
@@ -82,6 +82,37 @@ def get_staking_ledger(self):
 
         return ((h, o.value) for (h, o) in query)
 
+    # TODO: move to a separate file
+    """ accesses current era """
+    def current_era(self):
```
altonen, line 87: I think you can just simply return the result without storing it to a temporary variable

iljakuklic (approved):

TheQuantumPhysicist (approved):
Please check Aaro's comments and fix them later. He said here they're not critical.

PR #112, opened 2021-11-10
staking: reward the same amount to the block author's pallet-balance.

b-yap:
rewarding block authors both in the pallet-balance and utxo system.


altonen (commented):
Unit tests do not pass

b-yap:
needs further testing

PR #119, opened 2021-11-12
Roy/docs

royster57:
Updated the docs for transactions, tokens, and node start-up.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,9 +3,23 @@
 
 ## Quick start 
 Binaries can be found [here](https://github.com/mintlayer/core/releases).
+
+Running a node also requires as input a chain specification.
+Currently a single chain specification for the testnet is provided, and can be downloaded using curl:
+
+```
+curl --proto '=https' -sSf 								\
+    https://raw.githubusercontent.com/mintlayer/core/master/assets/Testnet1Spec.json 	\
+    --output Testnet1Spec.json
+```
 Download and run:
 ```
-mintlayer-core --base-path data/my_first_ml_node --validator --rpc-external --rpc-methods Unsafe --chain=Testnet1Spec.json
+mintlayer-core 				\
+    --base-path data/my_first_ml_node 	\
+    --validator 			\
+    --rpc-external 			\
+    --rpc-methods Unsafe 		\
+    --chain=Testnet1Spec.json
 ```
 
 to start a node. It will automatically connect to the Mintlayer bootnodes.
@@ -80,7 +94,7 @@ See [Mintlayer installation on Windows](windows_installation.md)
 ## Running a node
 Clone the repository:
 
-```bash
+```
 git clone https://github.com/mintlayer/core.git
 ```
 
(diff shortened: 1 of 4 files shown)


docs/tokens.md
```diff
@@ -1,25 +1,21 @@
...
 ## Transfer Tokens 
 
 **TODO sentence fragment, I don't understand...**
 **TODO When do we NOT use the TxData field?**
-For transfering funds to another person in a given UTXO. To send MLT we will use the MLT token ID, which is equal to 0. If the token ID is equal to the ID of the MLS-01 (**TODO what is MLS-01**) token, then the amount of token is transferred to the recipient. The commission is taken only in MLT (**TODO what is this commission**. If the token ID is equal to the ID of any NFT, then the data of this NFT is transferred to the recipient without changing the creator field. The UTXO model itself allows to determine the owner of the NFT.
+For transfering funds to another person in a given UTXO. To send MLT we will use the MLT token ID, which is equal to 0. If the token ID is equal to the ID of an MLS-01 token, then the amount of the token is transferred to the recipient. The commission is taken only in MLT (**TODO what is this commission**. If the token ID is equal to the ID of any NFT, then the data of this NFT is transferred to the recipient without changing the creator field. The UTXO model itself allows to determine the owner of the NFT **TODO Explain this**.
```
muursh, line 18: What do you mean by commission? Do you mean the transfer fee? As it stands we enforce a MLT tx fee as we have no way of helping a block producer select the most valuable tx fee when they're in random tokens. Is a bentoken worth more than a roytoken? Are they both worth less than MLT?

docs/README.md
```diff
@@ -80,11 +94,11 @@ See [Mintlayer installation on Windows](windows_installation.md)
...
 Clone the repository:
 
-```bash
-git clone https://github.com/mintlayer/core.git
+```
+git clone https://github.com/royster57/launch_docs.git
 ```
 
-or download the [zip](https://github.com/mintlayer/core/archive/refs/heads/master.zip).
+or download the [zip](https://github.com/royster57/launch_docs/archive/refs/heads/main.zip).
```
muursh, line 101: Do we want these elsewhere?

docs/transaction.md
```diff
@@ -2,17 +2,11 @@
 
 ## UTXO overview
 
-**TODO maybe system/model instead of structure?**
-Mintlayer uses the Bitcoin UTXO structure instead of the account-based models of Ethereum, Ripple, Stellar, and others.
-**TODO need explanation of this sentence, it is not clear to me**
-Since each transaction's output is stored separately (even when sent to a single address), it is only possible to spend the entire transaction’s output.
+Mintlayer uses the Bitcoin UTXO system instead of the account-based models of Ethereum, Ripple, Stellar, and others. There are three essential reasons for this:
```
muursh, line 5: I'd say Bitcoin-esque as it's not identical to Bitcoin's.

scripts/wallet.py
```diff
@@ -9,9 +9,12 @@
 import logging
```
muursh, line 9: I've heard about this file from Aaro so I won't comment :D

docs/transaction.md
```diff
@@ -2,17 +2,11 @@
...
+Mintlayer uses the Bitcoin UTXO system instead of the account-based models of Ethereum, Ripple, Stellar, and others. There are three essential reasons for this: 
 
-There are three essential reasons for choosing the UTXO model: 
-
-- It is compatible with technologies already implemented in Bitcoin, such atomic swaps and the Lightning Network.
+- The utxo model is compatible with technologies already implemented in Bitcoin, such atomic swaps and the Lightning Network.
  
-**TODO - why does this improve privacy?**
-- It is more privacy-oriented: a single wallet usually utilizes multiple addresses, making it difficult and sometimes impossible to determine which addresses belong to whichs user. 
+- The utxo model is more privacy-oriented: a single wallet usually utilizes multiple addresses, making it difficult and sometimes impossible to determine which addresses belong to which user.
```
muursh, line 9: I'd say `it's possible to utilise multiple addresses` rather than `usually utilises` while it might be true for a certain chunk of the population I have no idea if that claim holds for the wider world

docs/tokens.md
```diff
@@ -66,13 +62,21 @@ TO DO
 ## Wallet
 
 TO DO
-## Issue and Transfer Tokens
+## Issuing and Transferring Tokens
 
 **TODO who are these examples meant for?**
+**TODO It looks like alice is giving carl normal tokens?**
```
muursh, line 68: What do you mean by normal tokens? MLS-01 is the "basic token" standard on ML. If you issued a RoyToken it would likely be MLS-01, if you think of a basic token on Ethereum the chances are it's an erc-20 token, this is the closest thing to that on ML. We also have MLS-02 (well not yet we don't - they're confidential tokens) and MLS-03 which are NFTs. Given the code below is for MLS-01 I think it's fair to call them "normal tokens" in the sense they don't have any special properties for CTs or being NF. I'd stick to the MLS-01 nomenclature over "normal" though as changing the wording everywhere will get confusing.

docs/tokens.md
```diff
@@ -66,13 +62,21 @@ TO DO
...
+**TODO what is this Origin::signed business? Go to definiton takes me to the construct_runtime macro call** 
+**TODO should explain the index argument in tx.outpoint(1)**
+**TOOD: In the definition of MLT unit, it's 1000 * 100_000_000. Why written this way. One hundred billion**
```
muursh, line 71: A rip-off of LN. Bitcoin breaks down to 100_000_000 satoshis. LN introduces the millisatoshi (sat/1000) hence the weird notation of 1_000 * 100_000_000

docs/tokens.md
```diff
@@ -66,13 +62,21 @@ TO DO
...
+**TODO Does it cost 1000 MLT to issue a token?**
```
muursh, line 72: It shouldn't be 1000 MLT. The figures discussed so far have been 50 MLT and 100 MLT.

muursh: @sinitcin

docs/tokens.md
royster57: This was basically me confused and mumbling to myself. I meant coins. I didn't understand the TxOutput containing both a nonzero Value and a TokenIssuance data field.After talking with Sam I understood that you can transfer tokens and coins in the same transaction.

docs/transaction.md
royster57: Changed it.

scripts/wallet.py
royster57: XD

docs/transaction.md
royster57: Changed

docs/tokens.md
royster57: Ah!

docs/tokens.md
royster57: Anyhow as I understood from my chat with Sam, the 1000 MLT coins mentioned here are token Alice gives to Karl, and not a fee paid to mint the token. If I understand correctly this example does not take into account fees for minting the token.

muursh: Cool

PR #120, opened 2021-11-15
Updated docs for node start-up, tokens, and transactions

royster57:


docs/README.md
```diff
@@ -3,9 +3,23 @@
...
+```
 Download and run:
 ```
-mintlayer-core --base-path data/my_first_ml_node --validator --rpc-external --rpc-methods Unsafe --chain=Testnet1Spec.json
+mintlayer-core 				\
+    --base-path data/my_first_ml_node 	\
+    --validator 			\
+    --rpc-external 			\
+    --rpc-methods Unsafe 		\
+    --chain=Testnet1Spec.json
```
iljakuklic, line 22: I never used the `--rpc-external` and `--rpc-methods Unsafe` options so at least for basic usage, these are not necessary. The question is whether users should pass them by default. Ideally, the command to run the node should be as simple as possible.

zorvan (approved):

docs/README.md
TheQuantumPhysicist: If I understand correctly, the `rpc-external` flag binds to all network interfaces with IP 0.0.0.0, meaning that anyone can connect to the node from outside the network. And `rpc-methods Unsafe`, basically allows anyone connecting to the RPC to be an owner of the node, and hence, expose keys. Meaning: We're basically exposing everything to everyone and allowing anyone connecting to RPC to do whatever they want... what's the logic behind this?! Did we try not adding them and had problems?

royster57: This is the command to run as I understood it from Amin.
@zorvan  What do you think of Lukas's and Sam's comments?

zorvan: They are right. Latest version of cli switches for user and production id reaped off from those.

royster57: @zorvan  so what is the correct command? Should I simply remove --rpc-external and --rpc-methods Unsafe?

TheQuantumPhysicist: As in every issue in programming, that depends on what you want to do. The right solution, in my opinion, is to remove these and tell the user that they will be binding to `localhost`, hence, no connections from the outside will be possible and explain that this is a safety procedure and using tunnels is usually better than opening to the outside, and then explain that unsafe methods will not work unless the user enables them.

TheQuantumPhysicist: And let me explain that hacks with externally accessible RPC happened before: 

https://www.coindaily.co/2801-2/

docs/tokens.md
```diff
@@ -30,11 +36,19 @@ TxData {
...
 ```
 
-## Issue Tokens
-When issuing a new token, we specify the data for creating a new token in the transaction input, where the `token_id` is a hash of the inputs. **TODO which inputs?**
-**TODO explain remaining fields**
+## Issuing Tokens
+
+**TODO is the sentence below true? In the example, I don't see any data relating to the token in the input.** 
+
+When issuing a new token, we specify the data for creating a new token in the transaction input, where the `token_id` is a hash of the inputs.
```
sinitcin, line 43: When issuing a new token, we specify the data for creating a new token in the transaction **output**. After that, a wallet should calculate the `token_id` that is a hash of the first input of the issuance transaction. And use this token_id is able to send tokens to anyone.

docs/tokens.md
```diff
@@ -30,11 +36,19 @@ TxData {
...
+
+**TODO explain the remaining fields (token_ticker, metadata_URI)**
```
sinitcin, line 45: `token_ticker` is a short name of the token. No longer than 5 symbols.

`metadata_URI` is a web link to a JSON file where we can store additional information for the token: https://github.com/mintlayer/core/pull/67 Over here you can find the paragraph: `Format of data located by reference metadata_URI`

docs/tokens.md
```diff
@@ -30,11 +36,19 @@ TxData {
...
+
+**TODO the token issuance struct shown here has a token_id which does not appear in the example. In staging, this is commented out. Why?**
```
sinitcin, line 47: During the review process, Sam tell me that I have to remove token_id in TokenIssuanceV1 because it's a calculatable field from an of the first input. I will clear comments.

docs/tokens.md
```diff
@@ -1,25 +1,31 @@
...
-- Transfer Tokens or NFT
-- Token Issuance
-- Burning tokens
+2. NFTs
+
+A transaction involving transaction output carries a (possibly empty) `data` field specifying the purpose of the transaction.
+
+**TOOD when is there no data field?**
+
+The data field can be any of the fol
```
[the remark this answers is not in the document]
royster57: Thanks

docs/tokens.md
```diff
@@ -1,25 +1,31 @@
...
+1. *MLS-01 tokens*: MLS-01 is the "basic" Mintlayer token standard, analogous to, say, ERC-20 tokens on Ethereum.
 
-- Transfer Tokens or NFT
-- Token Issuance
-- Burning tokens
+2. NFTs
+
+A transaction involving transaction output carries a (possibly empty) `data` field specifying the purpose of the transaction.
+
+**TOOD when is there no data field?**
```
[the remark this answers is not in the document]
royster57: Yup

docs/tokens.md
royster57: Thanks, that clarifies things! I will modify the explanation.

docs/tokens.md
royster57: @sinitcin Which characters can be used in the token ticker? Are numbers, special characters allowed?

TheQuantumPhysicist (approved):

muursh (approved):

PR #123, opened 2021-11-15
Refactor staking-related queries to separate file

royster57:


test/functional/test_framework/mintlayer/staking.py
```diff
@@ -0,0 +1,42 @@
+import substrateinterface
+
+
+class Staking(object):
+    def __init__(self, substrate):
+        self.substrate = substrate
```
iljakuklic, line 6: I personally find it a bit overkill to define a new class just to add a couple of methods. What is the reasoning behind doing it this way?

b-yap: well, it was my idea, because the functions here are for `pallet-staking` storage, and has no utxo aspect of it.
A couple of those functions are:
`current_era` - returns current era of the chain
`withdrawal_era` - given a controller account, return the era which withdrawal is allowed.
`get_staking_ledger` - given a controller account, returns the `StakingLedger` with the ff info: the total staked, the stash account, if unlocking has been triggered

say for example, if we're doing a staking, it's not enough to check whether the `StakingCount` of the utxo system has been altered. The test should also reflect that the `pallet-staking` side was successful, like:
1.  check the `Bonded` storage, if the stash account and the controller account has been bonded successfully.
2. check whether another entry is added in the `StakingLedger`, with the controller account as the key;
3. The `Validators` storage has a new entry

iljakuklic: Ah I misread the code originally, thinking this inherited from `utxo.Client`. Makes more sense now. Disregard my original comment.

TheQuantumPhysicist (approved):

PR #124, opened 2021-11-15
script: Add some exhaustive testing

iljakuklic:


libs/chainscript/src/test/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        // Run the script and record mismatches between expected and actual outputs.
+        let result = run_script(&TestContext::default(), &script, vec![].into()).ok();
+        if expected != result {
+            println!("FAIL {:?}: {:?} vs. {:?}", script, result, expected);
+            fails += 1;
+        }
+    }
+
+    // Let the test fial if we have at least one mismatch.
```
muursh, line 54: fail not fial

libs/chainscript/src/test/mod.rs
```diff
@@ -0,0 +1,56 @@
+use crate::*;
+
+// Test the interpreter on all 4-byte combinations of non-trivial opcodes.
+#[test]
+//#[ignore = "This test is too expensive to run by default"]
```
muursh, line 5: Did you mean to comment the ignore?

iljakuklic: Not sure. It runs pretty quickly in release mode, but under debug, I will have to measure.

iljakuklic: Running this test with debug build takes 147 seconds on my laptop.

iljakuklic: What do we aim for in these tests? Are tests that run a long-ish time under debug acceptable or do we want to disable them by default (make opt-in)?

muursh: 147s is okay for me but is probably approaching the limit of "I'm getting bored waiting for this"-ness so maybe making them opt-in is a good idea. If someone touches the script stuff then they know to run them anyway so we're probably safe.

TheQuantumPhysicist (approved):

muursh (approved):

libs/chainscript/src/test/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+                    parts.map(|s| Vec::from_hex(s).expect("hex item").into()).collect();
+                Some(stack.into())
+            }
+            _ => unreachable!("bad format"),
+        };
+
+        // Run the script and record mismatches between expected and actual outputs.
+        let result = run_script(&TestContext::default(), &script, vec![].into()).ok();
+        if expected != result {
+            println!("FAIL {:?}: {:?} vs. {:?}", script, result, expected);
```
altonen, line 49: Just wondering would `eprintln!()` work better but I think that both `stdout` and `stderr` are always captured so not sure if it makes a difference

altonen (approved):

libs/chainscript/src/test/mod.rs
iljakuklic: In release mode, it is about 5s on my laptop btw

PR #126, opened 2021-11-16
disable some tests

royster57:
Not a real pull request, just for triggering the pipeline to determine which functional tests are not terminating.


TheQuantumPhysicist:
This is probably unnecessary. We can reopen it if we need it.

PR #127, opened 2021-11-16
runtime: Cargo feature to make block production faster

iljakuklic:
Add a compilation-time feature to shorten the block interval to 3 seconds. Intended mainly for testing.

This is the approach to adjusting block interval that I found easiest to implement. The PR is mainly intended as a discussion starter. Do we want a command-line switch (would require some hacks)? Is this sufficient for testing?

Use `cargo build --release --features fast` to enable 3s block times.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,9 +67,9 @@ jobs:
       with:
         submodules: recursive
     - name: Build
-      run: cargo build --release
+      run: cargo build --release --features short-block-time
     - name: Run tests
-      run: cargo test --release
+      run: cargo test --release --features short-block-time
     - name: Prepare functional tests
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,5 +84,5 @@ jobs:
       uses: actions/upload-artifact@v2
       if: failure()
       with:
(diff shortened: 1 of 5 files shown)


TheQuantumPhysicist:
OMG... I'm working on this RIGHT NOW... hehe.

Good on you. I just called the feature `short-block-time` instead. I don't recommend "fast" because it can mean many things.

iljakuklic:
> OMG... I'm working on this RIGHT NOW... hehe.
> 
> Good on you. I just called the feature `short-block-time` instead. I don't recommend "fast" because it can mean many things.

Changed the feature name to yours. Are our code changes now byte-identical? :smile:

TheQuantumPhysicist:
> > OMG... I'm working on this RIGHT NOW... hehe.
> > Good on you. I just called the feature `short-block-time` instead. I don't recommend "fast" because it can mean many things.
> 
> Changed the feature name to yours. Are our code changes now byte-identical? 😄

Yep, except that I chose 5 seconds instead of 3. I think the shorter the better though. 

I just tested this in the functional tests and this seems to work and the tests are 3x fast. I'll add a commit with CI changes and then we merge. Then we get everyone to rebase on this if it passes here.

TheQuantumPhysicist (approved):

PR #130, opened 2021-11-17
Make unlock_and_withdraw test deterministic

royster57:


test/functional/feature_staking_unlock_and_withdraw.py
```diff
@@ -87,9 +88,11 @@ def run_test(self):
         assert_equal(events[2].value['event_id'], 'StakeUnlocked')
 
         ledger = list(client.get_staking_ledger())
+
         assert_equal(len(ledger),2)
 
-        time.sleep(500)
+        while client.current_era() < client.withdrawal_era(alice):
+            time.sleep(1)
```
iljakuklic, line 95: I think there should be a hard timeout as well in case there is a bug somewhere that causes `withdrawal_era` to never be reached.

royster57: I agree. I will add it.

royster57:
@TheQuantumPhysicist Take a look and approve if you like it.

test/functional/feature_staking_unlock_and_withdraw.py
```diff
@@ -87,9 +88,16 @@ def run_test(self):
         assert_equal(events[2].value['event_id'], 'StakeUnlocked')
 
         ledger = list(client.get_staking_ledger())
+
         assert_equal(len(ledger),2)
 
-        time.sleep(500)
+        num_attempts = 0
+        max_attempts = 10000
```
TheQuantumPhysicist, line 95: 10000 seconds is like 3 hours... you sure?

test/functional/feature_staking_unlock_and_withdraw.py
```diff
@@ -87,9 +88,16 @@ def run_test(self):
...
+        while client.current_era() < client.withdrawal_era(alice):
+            num_attempts += 1
```
TheQuantumPhysicist, line 97: To make this in terms of time, I'd use a period instead of num_attempts, and then wait that period in the sleep. When max_wait is exceeded, I'd call it an error. Since Python is very tolerant with types, you can make use of it and have only time instead of time + unitless numeric attempts.

TheQuantumPhysicist (approved):
The wait mechanism can be improved. I made my comments there. But other than that, looks good. Good job!

test/functional/feature_staking_unlock_and_withdraw.py
royster57: @TheQuantumPhysicist I chose this a bit randomly.... Should we go for 20 minutes? This is twice as long as it *should* take:

TheQuantumPhysicist: Well, I'd prefer that you change this whole thing to wait for time units, not retry count. If it isn't clear what I mean, hit me with a DM.

test/functional/feature_staking_unlock_and_withdraw.py
royster57: @TheQuantumPhysicist Is "error" different from "Exception"?

PR #133, opened 2021-11-19
Clean up Substrate node template remains

iljakuklic:
* Remove mentions of node template, replace them with Mintlayer terminology
* Remove the `template` pallet. While this might be somewhat useful as a starting point for custom pallets, it is easy to find so no need to pollute the repo with it.


docker-compose.yml
```diff
@@ -2,16 +2,16 @@ version: "3.2"
...
-      - CARGO_HOME=/var/www/node-template/.cargo
+      - CARGO_HOME=/var/www/mintlayer-core/.cargo
     volumes:
-      - .:/var/www/node-template
+      - .:/var/www/mintlayer-core
       - type: bind
         source: ./.local
         target: /root/.local
-    command: bash -c "cargo build --release && ./target/release/node-template --dev --ws-external"
+    command: bash -c "cargo build --release && ./target/release/mintlayer-core --dev --ws-external"
```
iljakuklic, line 17: I have not tested the docker changes here, it is just a blind search/replace. Is anybody using this?

muursh: I don't usually use it but I can give it a quick spin as I have played with it before

muursh: Seems fine to me

muursh (approved):
Everything runs fine locally so looks good to me

docker-compose.yml
altonen: Somewhat unrelated but at some point we probably should provide a Dockerfile that installs everything needed and launches the node. Right now a lot of people seem to have problems connecting to our testnet but if all they had to do was run `docker run -t mintlayer-core`, I'm sure the number of support requests would drop

altonen (approved):

docker-compose.yml
muursh: Yeah dealing with docker support requests will be much more fun :smiley:

PR #135, opened 2021-11-22
staking functional tests: expect exception on testing failed transactions

b-yap:
also fixing transaction with fee exceeding `u64::MAX`

### DIFF
--- a/test/functional/feature_alice_bob_test.py
+++ b/test/functional/feature_alice_bob_test.py
@@ -17,6 +17,7 @@ from test_framework.util import (
     connect_nodes,
     wait_until,
 )
+from test_framework.messages import COIN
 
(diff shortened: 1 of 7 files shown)


sinitcin (approved):

muursh (approved):

TheQuantumPhysicist (approved):

iljakuklic (approved):

altonen:
I think the tests may still fail because programmable pool tests send two invalid transactions to verify that the smart contract is not accepted and those asserts are not addressed in this PR. If people are OK with it, it may be easier to just disable smart contract tests in this PR and once this is merged, I can send a PR that properly fixes the PP tests.

